PR #2, opened 2021-08-31
update yml file - try paths for triggering CI for unit test

Boopster:


.github/workflows/ci.yml
```diff
@@ -3,9 +3,9 @@ name: Continuous Integration
 # On every pull request on the develop branch, execute all tests [CI]
 on:
   pull_request:
-    branches:
-      - develop
```
stephenbit, line 7: Should we remove `develop` now that it has been deleted and we aren't using it?

Boopster: I think you have clicked on the file showing the deleted lines

.github/workflows/ci.yml
```diff
@@ -3,9 +3,9 @@ name: Continuous Integration
 # On every pull request on the develop branch, execute all tests [CI]
 on:
   pull_request:
```
mikemherron, line 5: Think this looks good - I'm not super familiar with GitHub actions, but would it be simple to update this to also run all tests on every commit, as well as on each pull request? Most times I've worked with CI in the past it's usually been on each commit.

Boopster: I will have a look.

PR #4, opened 2021-09-04
Feature/sva 61 add node server project

Boopster:


mikemherron (approved):
Only small question was should the ".expo-shared" folder still be at the top, or should it also be moved down to the app folder?

Boopster:
> Only small question was should the ".expo-shared" folder still be at the top, or should it also be moved down to the app folder?

Good point @mikemherron - I will move and resubmit

gunnx:
I would consider having the node application in its own repo

PR #7, opened 2021-09-08
SVA-47_Investigate_Jira_api

hqureshi88:
Just a practise run on pull request


mikemherron:
This looks good 👍 exactly what we needed, just a super basic code sample showing how call the Jira API. Since the code at this point is just "POC", it's worth checking it in but possibly not being returned from the actual URL we're going to use - the reason being that ideally we'd want to do things in this sort of order:

1. Define the response format for the API (https://sta2020.atlassian.net/browse/SVA-60)
2. Add a stub response from the API that looks like the real response (https://sta2020.atlassian.net/jira/software/projects/SVA/boards/145?selectedIssue=SVA-63)
3. App client starts to use this response to render project list
4. The API can then be updated to return the real response with minimal chances

Since there is a PR already up for 2, if we were to check this in we wouldn't be able to do 3, as the API would be returning the JIRA list in a format we don't want at the moment - so wondering if it's possible to have this code exist somewhere in the project as a simple command line node script?

gunnx:
suggest looking at using https://axios-http.com/ for making requests once you start building API calls for real

hqureshi88:
Hi everyone, 

Please have a look at my code. (yes yes...I stupidly pushed it with token but luckily Atlassian security detected this and revoked it from my jira account!). The code now fetches the relevant JIRA projects, which now displays below, 

Snapshot JIRA api

I will need to liaise with Cat to see what we can do with the fields shown as 'customfield_10112' etc as key - my thoughts would be to investigate JIRA documentation with Cat and see if we could change the name of these fields to meaningful names, which would help smooth the process of extracting information via our api and return data to be consumed by the browser.

Util/api_definition.json
```diff
@@ -0,0 +1,45 @@
+{
```
mikemherron, line 1: I don't think this is quite  matching the required format for an Open API def - I had a look at it and made some changes:

* add in proper path structure
* add openapi version to top
* add operation description and response code
* Update the skills definition to an array of strings

I also added an example based on @sharkand73 mock data, as well as move the schema in to a seperate object and referenced (I know we said we'd leave this for now but I was using a validator and it was complaining). Have put the updated def below, will let you take a look and let me know if you want to chat through it:

```
{ 
	"openapi" : "3.0.3",
    "info": {
        "title": "STA Projects API",
		"description" : "API to list and query available STA volunteer projects",
        "version": "1.0"
    },
	"paths": {
		"/projects": {			
				"get" : {
					"description" : "Returns a list of available projects",
					"operationId" : "get_projects",
					"responses" : {
						"200" : {
							"description" : "OK",
							"content":{
								"application/json": {
									"example" : [
										{
											"key": "PKAVS",
											"name": "project-website-pkavs",
											"type": "Team-managed software",
											"client": "PKAVS",
											"role": "QA Tester",
											"description": "Blah blah",
											"skills": ["Acceptance Testing"],
											"hours": "5-10 hours per week",
											"required": "One person",
											"buddying": false
										},
										{
											"key": "SVA",
											"name": "STA-volunteer Apps",
											"type": "Team-managed software",
											"client": "STA - internal",
											"role": "Lead developer",
											"description": "Blah blah",
											"skills": ["React Native", "Node.js"],
											"hours": "5-10 hours per week",
											"required": "One person",
											"buddying": true
										}
									],
									"schema" : {
										"$ref": "#/components/schemas/project" 
									}
								}
							}
						}
					}
				}
			
		}       
    },
	"components" : {
		"schemas" : {
			"project" : {
				"type": "array",
				"description" : "Describes a project available for volunteering",
				"items": {
					"type": "object",
					"properties": {
						"key": {
							"type": "string",
							"description": "Unique identifier for the project"
						},
						"name": {
							"type": "string",
							"description": "Descriptive name for the project"
						},
						"type": {
							"type": "string",
							"description": "The type of project"
						},
						"client": {
							"type": "string",
							"description" : "The name of the client"
						},
						"role": {
							"type": "string",
							"description" : "The role the volunteer is expexcted to play on the project"
						},
						"description": {
							"type": "string",
							"description" : "Description of what the project work involves"
						},
						"skills": {
							"type": "array", 
							"description" : "List of skills required for the project",
							"items" : {
								"type" : "string"
							}
						},
						"hours": {
							"type": "string",
							"description" : "Hourly commitment required for the project"
						},
						"required": {
							"type": "string",
							"description" : "Number of people required"
						},
						"buddying": {
							"type": "boolean",
							 "description" : "Wether the project supports buddying"
						}
					}
				}
			}			
		}
	}
}
```

Util/script.js
```diff
@@ -0,0 +1,33 @@
+import express from "express";
```
mikemherron, line 1: It might be worth moving this file in to "api/util", and perhaps moving the schema def in to "api/definition" ? Would also keep the folder name lowercase for consistency.

api/app.js
```diff
@@ -1,11 +0,0 @@
-import express from "express";
```
mikemherron, line 1: Did you delete this by accident? We should probably keep this file here, as once we merge to main @sharkand73 dummy data response will be returned.

mikemherron (changes requested):

api/app.js
hqureshi88: Ok, managed to bring back the file i deleted

Boopster (approved):
Approved - all looking good when run locally

mikemherron (approved):

PR #22, opened 2021-09-12
Bug/sva 65

Boopster:


mikemherron (changes requested):
I think we might need to add api/node_modules to gitignore? Looks like they have been checked in for the API folder.

Boopster:
> I think we might need to add api/node_modules to gitignore? Looks like they have been checked in for the API folder.

I think this is now resolved.

mikemherron (approved):

CatMac228 (approved):
Happy to approve :)

CatMac228 (approved):

PR #40, opened 2022-01-24
added a endpoint to get a single event as well as adding tests

stevenjburrows:
- [Description](#description)
- [Checklist](#checklist)

# Description

- Please include a summary of the change and which issue is fixed.
- List any dependencies that are required for this change.

Fixes # (issue number)

## Type of change
Added new endpoint to get a single event from AirTable

- [x] New Feature (non-breaking change which adds functionlity)

# Checklist

- [] I have commented my code, pariculary in hard-to-understand areas


CatMac228 (approved):

api/__tests__/events.test.js
```diff
@@ -6,7 +6,7 @@ const nock = require('nock')
 
 axios.defaults.adapter = require('axios/lib/adapters/http')
 
-const scope = nock('http://localhost:5000')
+const scopeEvents = nock('http://localhost:3000')
     .get('/airtable/events')
```
joefhall, line 10: Not a biggie, but wonder if it might be useful to make all the events endpoints begin simply `/events` rather than `/airtable/events`?  Then if we changed the source of the events data later so it's no longer coming from AirTable, the URL wouldn't need to be updated here or in the app.  Also maybe it's good for anything consuming this endpoint not to have to care where the data's coming from?  All they need to know is what thing (e.g. events) they want.  Totally appreciate there may be other good reasons why you did this though.

api/__tests__/events.test.js
```diff
@@ -38,19 +38,77 @@ const scope = nock('http://localhost:5000')
         "createdTime": "2021-09-30T17:41:59.000Z"
       }])
 
+      const scopeSingleEvent = nock('http://localhost:3000')
+      .get('/airtable/event/recAb4kAFdkeKKnLU')
+      .reply(200, 
+        {
+          "id": "recAb4kAFdkeKKnLU",
+          "fields": {
+            "Event Name": "Showcase - Volunteer app",
```
joefhall, line 47: Maybe this isn't a priority at this stage or would be a pain as some of the AirTable column names contain spaces and capitals, but I wondered if we should use a consistent case in the event JSON object?  (not just in the test but I spotted it here) -- e.g. `id`, `createdTime`, etc are camel case but `Event Name`, etc aren't.  Don't want to hold things up so maybe not essential, but it's something that struck me seeing the test data.

joefhall (approved):

api/__tests__/events.test.js
stevenjburrows: That is a possibility, I added the airtable incase something other than evens were added

api/__tests__/events.test.js
stevenjburrows: This is a copy of how Airtable returns the results.  I just copied and pasted the JSON from insomnia, it might need changed on the airtable itself

PR #50, opened 2022-02-24
Feature/SVA-137 | wire backend to frontend

PammyLo:
- [Description](#description)
- [Checklist](#checklist)

# Description

- This wires the project list in the app to the back-end API, where it was using seed data. 

Fixes [SVA-137](https://sta2020.atlassian.net/browse/SVA-137)

## Type of change

- [X] Bug fix (non-breaking change which fixes an isssue)
- [X] New Feature (non-breaking change which adds functionlity)
- [] Breaking change(fix or feature that would cause existing functionlity to not work as expected)

# Checklist

- [] I have commented my code, pariculary in hard-to-understand areas

### DIFF
--- a/app/src/Containers/HomeContainer.tsx
+++ b/app/src/Containers/HomeContainer.tsx
@@ -36,16 +36,16 @@ const ProjectList: FC<ProjectProps> = ({ data }) => {
 }
 
 const HomeContainer = () => {
-  const [fetchAll, { data: listData }] = useLazyFetchAllQuery()
+  const [fetchAll, { data: projects }] = useLazyFetchAllQuery()
 
   useEffect(() => {
     fetchAll('')
   }, [fetchAll])
 
-  if (listData) {
+  if (projects) {
     return (
       <Theme>
-        <ProjectList data={listData} />
+        <ProjectList data={projects} />
       </Theme>
     )
   } else {
--- a/app/src/Containers/HomeContainer.tsx
+++ b/app/src/Containers/HomeContainer.tsx
@@ -36,16 +36,16 @@ const ProjectList: FC<ProjectProps> = ({ data }) => {
 }
 
 const HomeContainer = () => {
-  const [fetchAll, { data: projects }] = useLazyFetchAllQuery()
+  const [fetchAll, { data: listData }] = useLazyFetchAllQuery()
 
   useEffect(() => {
     fetchAll('')
   }, [fetchAll])
 
-  if (projects) {
+  if (listData) {
     return (
       <Theme>
-        <ProjectList data={projects} />
+        <ProjectList data={listData} />
       </Theme>
     )
   } else {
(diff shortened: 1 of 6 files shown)


PammyLo:
Screenshot 2022-02-24 at 15 38 00

app/src/Services/modules/projects/index.ts
```diff
@@ -1,24 +1,31 @@
...
+    description: string
+    candidateCoreSkills: string
+    candidateTime: string
+    suitableForBuddy: boolean
+  },
+  ItData: {
+    projectName: string
+    charityName: string
+    charityVideo: ImageURISource
+  }
```
jppferguson, line 28: I think this is probably how the data for projects is coming through from the API, but wondering if we can get the backend folks to flatten these 2 objects into a single object rather than having these 2 nested `ResData` and `ItData` objects so it would be more like this?
```suggestion
res_id: string
projectType: string
jobRole: string
description: string
candidateCoreSkills: string
candidateTime: string
suitableForBuddy: boolean
projectName: string
charityName: string
charityVideo: ImageURISource
```

That way, when we access the project data (in `app/src/Components/Project/ProjectSummary.tsx`) we don't need to dig 2 levels deep into the object:
`<ProjectHeading title={project.projectName} />`
instead of:
`<ProjectHeading title={project.ItData.projectName} />`

We could maybe add a ticket in the backlog to look at doing the above separately?

app/src/Containers/HomeContainer.tsx
```diff
@@ -36,16 +36,16 @@ const ProjectList: FC<ProjectProps> = ({ data }) => {
...
   useEffect(() => {
     fetchAll('')
   }, [fetchAll])
 
-  if (projects) {
+  if (listData) {
     return (
       <Theme>
-        <ProjectList data={projects} />
+        <ProjectList data={listData} />
```
jppferguson, line 48: What was the reason for renaming this? If it was to make it clear that this is returning a list of projects, it might be better named something like `projectListData`?

jppferguson (commented):
Also as we spoke about it might be useful to add in the Readme that the new environment variables `EMAIL` and `API_KEY` are needed to run the API. That might be one for another ticket though as it's not really related to this PR

app/src/Services/modules/projects/index.ts
PammyLo: Good point, it would be neater that way. As discussed in the team meeting this morning a new ticket will be raised for this.

app/src/Containers/HomeContainer.tsx
PammyLo: Just tested it and it works when I change it back! I changed it to the final result in the backend because the data wasn't coming.. However, I think the real solution was the changes made in 'fetchAll'.
I've changed it back, thanks Jamie well spotted.

jppferguson (approved):
Looks good Pam! 🎉

DougieWougie (approved):
Looks great!

PR #54, opened 2022-03-07
feature/SVA-132 Speed up projects API endpoints

joefhall:
- [Description](#description)
- [Checklist](#checklist)

# Description

- This speeds up the API response from the `/projects` and `/projects/single` API end points
- It was taking some time to hit the Jira API each time someone makes a request to our API
- Now instead, IT/projects and resources results from the Jira API are saved in AirTable tables (this can be run as a scheduled task or on-demand)
- When the API endpoints are called, we can get data quickly from the AirTable tables, format and return it
- The data returned has changed format but it now follows the format defined in `/api/definitions/api_definition.json` (with slight suggested update to this), so it should be more consistent/predictable from now on for the front-end app

Ticket: [SVA-132](https://sta2020.atlassian.net/jira/software/projects/SVA/boards/145?selectedIssue=SVA-132)

# Testing notes

- You will need to run `npm install` from within the /api directory to install new dependencies.
- You will need to update/add environment variables in your `/api/.env` file -- please see `/api/.env.example` for what values you need (including what to use for the AirTable table names).  Please DM me on Slack if there are any other details you need and don't have.  Please also note that the Jira variables have been renamed -- you'll need to update these.  After changing any environment variables, you'll need to restart your local Node server (terminate the current process if you're running it, then do `npm run start` again).

## Type of change

- [x] This change requires a documentation update

# Checklist


api/.env.example
```diff
@@ -0,0 +1,7 @@
+JIRA_EMAIL = ""
+JIRA_API_KEY = ""
```
joefhall, line 2: While adding this file, I've suggested updating the names of the two Jira environment variables, to make this clearer and more future-proof (as we might want to add more email/API key variables for other services).

api/__tests__/events.test.js
```diff
@@ -1,178 +0,0 @@
-const request = require('supertest');
```
joefhall, line 1: This file has just moved to `/api/__tests__/routes`

api/__tests__/projects.test.js
```diff
@@ -1,150 +0,0 @@
-const app = require('../app');
```
joefhall, line 1: This file has moved to `/api/__tests__/routes`

api/.env.example
DougieWougie: Makes sense.

DougieWougie (changes requested):
Looking good Joe - there's a conflict with api/routes/projects.js that needs resolving though.

Pulled and ran the code without issue on Android.

DougieWougie (approved):
Great work Joe!

joefhall:
Thanks @DougieWougie!  Appreciate you looking at this and all your input

PammyLo (commented):
Apps works as expected in simulators and data loads in less than two seconds in Insomnia. 
Nice one @joefhall!

hqureshi88:
Api runs much faster and is now displaying combined data from both sources; resource board and IT board. Also the code is cleaner, readable & well structured to enable any future modifications - great job Joe 👯

hqureshi88 (approved):
Approving changes.

joefhall:
Thanks a lot @PammyLo @stevenjburrows @hqureshi88 for reviewing this too 👍

PammyLo (approved):

PR #71, opened 2022-06-24
Add tech stack / languages quick search

PammyLo:
- [Description](#description)
- [Checklist](#checklist)

# Description

- This adds functionality to the quick search buttons in the category 'Tech Stack / Languages'.
- It searches the 'skills' sections of the projects. Please note that the results for example for Java will also find JavaScript and React will find React Native, due to the skills being presented in a descriptive paragraph.
- I have installed @types/styled-components-react-native to resolve the missing declaration for styled-components/native I was getting in VS Code. (NPM install will be required in the /app folder for anyone with the same warning of a  'problem' in their IDE - note; the code ran fine, but a problem was highlighted by VS Code.   

Fixes # ([SVA-43](https://sta2020.atlassian.net/browse/SVA-43))

## Type of change

- [] Bug fix (non-breaking change which fixes an issue)

# Checklist


app/src/Containers/SearchContainer.tsx
```diff
@@ -307,7 +307,7 @@ const SearchContainer = () => {
         const fuse = new Fuse(projects, {
           keys: searchKeys,
           minMatchCharLength: 2,
-          threshold: 0.4,
+          threshold: 0.3,
```
joefhall, line 310: Just thinking about the conversation at the retro, could you add a comment here to say what `threshold` means and why it's set to 0.3?

joefhall (changes requested):
Thanks @PammyLo -- looks good and mostly worked well when I tested it (tricky to test it loads as there aren't many projects).

I did get a false positive when choosing 'React' (it fuzzy matched one of the words in the Datakirk project skills), but that's an issue with the data rather the code, I don't know that you could do more here.  Could we have a conversation with Joanna at the next stand-up about whether the skills field could be standardised when people are inputting into it (e.g. multiple choice)?

app/src/Containers/SearchContainer.tsx
PammyLo: Good call Joe!

PammyLo: The threshold was reduced because of the unwanted results, like the one you had for 'React'. I wonder if we should be using fuzzy search here at all? Agree this would be resolved if the skills field was standardised. Best to discuss that with Joanna.

joefhall: Yeah that's a good point @PammyLo -- looking at it again, maybe for the tech stack quick search it shouldn't use the fuzzy search and just look for the exact text instead, as these (Python, JavaScript, etc) are such specific terms.

joefhall: In fact, I can't remember a good reason why the quick search buttons need to use fuzzy search, so could generally simplify the quick search functionality and look for exact rather than fuzzy matches?  

(The benefit of fuzzy search is when people might misspell / write a word a little differently so it's useful for the free text search, but with the quick search buttons it's different - we set the search terms.  As long as text in the projects themselves is spelt the way we expect, an exact search could work here?)

PammyLo: The fuzzy search is needed for the roles quick search and best avoided for the other quick searches. 
Thanks @joefhall for the comments for improvement and a huge thanks for our joint session to analyse what was needed here!

PammyLo:
I have installed @types/styled-components-react-native to resolve the missing declaration I was getting for styled-components/native. NPM install will be required in /app for anyone with the same 'problem' warning in their IDE (it ran fine, but the problem was highlighted by VS Code).

app/src/Containers/SearchContainer.tsx
```diff
@@ -294,7 +297,33 @@ const SearchContainer = () => {
...
+    searchField: 'client' | 'description' | 'name' | 'role' | 'skills' | 'sector',
+  ): Projects => {
+    let results = [] as Projects
+
+    if (projects) {
+      results = projects.filter(
+        project => {
+          const anySearchQueriesMatching = searchQueries.some(searchQuery => {
+            if (typeof project[searchField] === "string") { // most fields are strings, but some are an array of strings (i.e. skills)
+              const isString = project[searchField] as string
```
joefhall, line 311: Just a small thing, but there's [a convention](https://www.samanthaming.com/tidbits/34-better-boolean-variable-names/) to often use `is` at the beginning of boolean variables, so it might be better to name `isString` something like `stringField` or `stringValue` instead

app/src/Containers/SearchContainer.tsx
```diff
@@ -241,31 +240,35 @@ const SearchContainer = () => {
...
     let searchQueries = [] as string[]
+    let results = [] as Projects
+
+    searchQueries.push(searchQueryChoice)
 
     if (searchField === 'role') {
       const relatedRoles = getRelatedRoles(searchQueryChoice)
 
       if (relatedRoles?.length) {
         searchQueries = relatedRoles
```
joefhall, line 257: Now that we've moved `searchQueries.push(searchQueryChoice)` up to line 251, `searchQueries` is no longer empty when we get to this line here, so we need to add to it rather than overwriting it.  So I think we should change this line here to `searchQueries = [...searchQueries, ...relatedRoles]` or `searchQueries = searchQueries.concat(relatedRoles)` (whichever you think is clearer)

app/src/Containers/SearchContainer.tsx
```diff
@@ -241,31 +240,35 @@ const SearchContainer = () => {
...
       }
+      results = fuzzySearchByArray(searchQueries, [searchField]) // we need to use fuzzy search as the roles should include various options of a role (i.e. without it developer does not include web developer results)
```
joefhall, line 259: It would be great if we can make this a little clearer - I think the core of the reason for using fuzzy search here is that role names are not exact (different charities use plenty of different ways of naming them), unlike tech stacks or sectors where it's reasonable to assume we can look for an exact match

app/src/Containers/SearchContainer.tsx
joefhall: Now that we're no longer adjusting how fuzzy search works to accommodate the skills quick search, I think it might be best if we leave this as it was at `0.4` as that setting seemed to work ok for other fields

joefhall (changes requested):
Nice one @PammyLo -- thanks for persevering with it!  Just a few small things I spotted

joefhall:
> I have installed @types/styled-components-react-native to resolve the missing declaration I was getting for styled-components/native. NPM install will be required in /app for anyone with the same 'problem' warning in their IDE (it ran fine, but the problem was highlighted by VS Code).

Def good to add this -- it's for TypeScript to work properly checking our code, thanks

app/src/Containers/SearchContainer.tsx
PammyLo: Quite right, thanks Joe

app/src/Containers/SearchContainer.tsx
PammyLo: Thanks Joe! I've changed it to concat, using the parameter searchQuery rather than relatedRoles as it's not for searchField 'roles'. Hope that makes sense! It gets the desired results so that's good ;)

app/src/Containers/SearchContainer.tsx
PammyLo: Thanks Joe, hope it's more clear now.

PammyLo:
All comments and change requests have been addressed. (Not sure why it still says 1 change requested.)

app/src/Containers/SearchContainer.tsx
joefhall: Sorry Pam, I think I wasn't clear enough with my comment --

It was ok keeping line 251 as `searchQueries.push(searchQueryChoice)` -- this adds the quick search button choice (e.g. `'Web Developer'` or `'Scrum Master'`) to the `searchQueries` array.
(NB this should be `searchQueryChoice` which is the text on the quick search button, not `searchQuery` which is the text in the free search box.  Feel free to rename these variables if you want -- maybe they're not clear enough at the moment.  Also we don't want to use `.concat` here as that merges an array with another array, whereas we're adding a single item to the array hence using `.push`)

It's this line 254 which needs to be `searchQueries = [...searchQueries, ...relatedRoles]` or `searchQueries = searchQueries.concat(relatedRoles)` so that we add any related roles to the `searchQueries` array, not replace it entirely.

With the code as you have it right now, the `searchQueries` array is being overwritten on line 254, so it no longer includes the `searchQueryChoice` string e.g. `'Web Developer'` or `'Scrum Master'`.  We'll probably still mostly get the right results, but there are cases where we won't -- for example with `'Scrum Master'` where there are no related roles for that choice (so line 254 as you currently have it would make `searchQueries` an empty array in that case).  If we keep `searchQueryChoice` as part of the `searchQueries` array it should still work.  (Appreciate this might be hard to test at the moment as we don't have projects matching all the different roles.)

Hope that makes some sense(!) 🙂  Happy to chat through more if that's helpful.

app/src/Containers/SearchContainer.tsx
joefhall: Are we ok to leave this as `0.4` instead of changing it to `0.3`?  I think we were only going to change this here because we were going to use fuzzy search for the skills / tech stack (which we're not doing now)

joefhall (changes requested):
Thanks Pam!  Just spotted a couple of things -- hope clear, but happy to chat if they're not, apologies if comments before weren't clear enough

M4RC1N76:
Pamela, I have been choosing a different skills options from the Tech Stack / Languages in the Search Options and I have noticed that when I pressed Java, the first result for the skills is Developer - Python. I was getting the same list of results in the same order when I pressed React Native, React and AWS. Is it intentional that I would get the general list of all skills just for now (during the time when the Volunteer App is under development), could you advise on that, please?

app/src/Containers/SearchContainer.tsx
PammyLo: sure, good idea, thanks Joe.

app/src/Containers/SearchContainer.tsx
```diff
@@ -294,7 +297,33 @@ const SearchContainer = () => {
...
+      results = projects.filter(
+        project => {
+          const anySearchQueriesMatching = searchQueries.some(searchQuery => {
+            if (typeof project[searchField] === "string") { // most fields are strings, but some are an array of strings (i.e. skills)
+              const stringSearchField = project[searchField] as string
+              return stringSearchField.toLowerCase().includes(searchQuery.toLowerCase())
+            }
+            else if (Array.isArray(project[searchField])) { // assume it's an array, ie skills
+              const arrayOfStrings = project[searchField] as string[]
+              return arrayOfStrings.some(item => item.toLowerCase().includes(searchQuery.toLowerCase()))
```
rc-1988, line 316: If you use `.includes` on this line, you'll get results for JavaScript when you search for Java, which probably isn't what you want. Since this method is meant to be checking for an exact match, then we could just compare for string equality instead of includes.

You might want to keep the `.toLowerCase` conversions on both sides of the equality, in order to catch e.g. JavaScript vs Javascript - depends on how much consistency we can expect from the source data on that front :-) (Might want a .trim() too if you think there might be trailing spaces?)

app/src/Containers/SearchContainer.tsx
```diff
@@ -241,31 +240,35 @@ const SearchContainer = () => {
...
 
+  
   const handleQuickSearchSubmit = (
     searchField: 'client' | 'description' | 'name' | 'role' | 'skills' | 'sector',
     searchQueryChoice: string,
   ) => {
     let searchQueries = [] as string[]
+    let results = [] as Projects
+    
+    searchQueries = searchQueries.concat(searchQuery)
```
rc-1988, line 251: Should this be 
```javascript
searchQueries = searchQueries.concat(searchQueryChoice)
```
?

app/src/Containers/SearchContainer.tsx
```diff
@@ -294,7 +297,33 @@ const SearchContainer = () => {
...
+  const exactSearchByArray = (
+    searchQueries: string[],
+    searchField: 'client' | 'description' | 'name' | 'role' | 'skills' | 'sector',
+  ): Projects => {
+    let results = [] as Projects
+
+    if (projects) {
+      results = projects.filter(
+        project => {
+          const anySearchQueriesMatching = searchQueries.some(searchQuery => {
```
rc-1988, line 309: Just a style thing: booleans usually have names starting with `is`, so a name like `isMatch` would be a helpful indicator that it's a boolean.

Or alternatively, since you're not doing anything with the boolean other than returning it, you could just return it straight away without giving it a name first: return searchQueries.some(...)

app/src/Containers/SearchContainer.tsx
PammyLo: Ah, that makes way more sense! After replacing the line yesterday there must have been a cache error as it worked fine. Thanks a million for checking, have amended and tested on both android and ios.

app/src/Containers/SearchContainer.tsx
PammyLo: Good call Rosie, thanks!

app/src/Containers/SearchContainer.tsx
PammyLo: No, this is the line causing the errors @M4RC1N76 has caught in testing. I did test but must have had a cache kick in or something (also bit tired last night). All resolved now. 

Thanks for catching this!

app/src/Containers/SearchContainer.tsx
PammyLo: Very true Rosy, good point. At the moment though the skills are not actually a dropdown in which matches as exact as that can be found. The skills are hidden in a paragraph. From conversations with Joanna it appears this will likely stay like that. This means a search for an exact comparison match will not return enough of the wanted results. This way the search will find the "exact" search word if it is in the paragraph but you are right, it does mean a search for example for Java will also return JavaScript and React will also return React Native, but considering the goal of the quick searches we felt this was a good solution. Happy to discuss tomorrow if you want and open to other solutions of course.

PammyLo:
Thanks for reviewing @M4RC1N76. You are right, that is not what we want! I had tested it but it must have been a cache (also very tired last night). It has now been fixed and is tested in Android and iOS. 
(Please note that the results for example for Java will also find JavaScript and React will find React Native, due to the skills being presented in a descriptive paragraph.)

M4RC1N76:
Thank you for clarifying Pamela. The Volunteer App is my first this massive size app,  so I feel like I need to ask about everything. Thanks again. 

Sent from Yahoo Mail on Android 
 
  On Thu, 4 Aug 2022 at 20:11, Pamela ***@***.***> wrote:   


@PammyLo commented on this pull request.

In app/src/Containers/SearchContainer.tsx:
> +    searchQueries: string[],
+    searchField: 'client' | 'description' | 'name' | 'role' | 'skills' | 'sector',
+  ): Projects => {
+    let results = [] as Projects
+
+    if (projects) {
+      results = projects.filter(
+        project => {
+          const anySearchQueriesMatching = searchQueries.some(searchQuery => {
+            if (typeof project[searchField] === "string") { // most fields are strings, but some are an array of strings (i.e. skills)
+              const stringSearchField = project[searchField] as string
+              return stringSearchField.toLowerCase().includes(searchQuery.toLowerCase())
+            }
+            else if (Array.isArray(project[searchField])) { // assume it's an array, ie skills
+              const arrayOfStrings = project[searchField] as string[]
+              return arrayOfStrings.some(item => item.toLowerCase().includes(searchQuery.toLowerCase()))

Very true Rosy, good point. At the moment though the skills are not actually a dropdown in which matches as exact as that can be found. The skills are hidden in a paragraph. From conversations with Joanna it appears this will likely stay like that. This means a search for an exact comparison match will not return enough of the wanted results. This way the search will find the "exact" search word if it is in the paragraph but you are right, it does mean a search for example for Java will also return JavaScript and React will also return React Native, but considering the goal of the quick searches we felt this was a good solution. Happy to discuss tomorrow if you want and open to other solutions of course.

app/src/Containers/SearchContainer.tsx
rc-1988: Thanks, this makes sense after the discussion this morning - think I'd been taking the 'exact search' description a bit too literally :-) Not sure what a better name would be: includesSearch, maybe?

PammyLo: I changed the name of the 'exactSearchByArray' to 'searchByArray' and added a comment to the .includes line to clarify.

rc-1988:
@M4RC1N76 would you like to review this one?

M4RC1N76:
Thanks for an update Pamela, that make sense. The description of the search 'searchByArray' is better than very specific 'exactSearchByArray'. As to the search result it works perfect!! While I was testing the Volunteer App, I have tried to choose AWS and there was no projects including AWS, hence I have received the message 'sorry we couldn't find any projects matching...'. Also I have received nothing for Java and JavaScript, due to lack of such projects at the moment. Next, I have tried Python option and received a project for Python.  Great work.

M4RC1N76 (approved):
All checked. The details are in the comments.

joefhall (approved):
Nice one @PammyLo and well done on sticking with it!

rc-1988 (approved):
Well done @PammyLo !! 👏

PR #72, opened 2022-07-01
SVA-170 | Stop API returning dummy projects data if it hits an error, remove unused files

joefhall:
- [Description](#description)
- [Checklist](#checklist)

# Description

- We had previously set up the projects API route to return dummy data if it ran into an error (this was because the error was usually that a new developer on the team didn't have the right .env details, this was a quick fix so that team members could get up and running quickly)
- But we don't want the API to be able to return dummy data when it's starting to be used with real app users, and we can help new team members get set up properly in other ways so they shouldn't need this
- Have tweaked error messages that the API returns to make sure these are helpful
- Have also taken the opportunity to remove a few API files that are no longer in use (some related to sample data)

Fixes [SVA-170](https://sta2020.atlassian.net/jira/software/projects/SVA/boards/145?selectedIssue=SVA-170)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# Checklist


DougieWougie (commented):
No issues!

joefhall:
Thanks @DougieWougie! 👍

PammyLo (approved):
Excellent @joefhall, thanks for the thorough tidy up!

joefhall:
Gracias @PammyLo!

PR #81, opened 2022-07-23
Feature/expand and reorganise readme

rc-1988:
- [Description](#description)
- [Checklist](#checklist)

# Description

A consolidation and tidyup of the changes on feature/expand_readme, in which we do the below. I don't seem to be able to tag him, but shoutout to Marcin: this was very much a collaborative effort!

- Expand the setup and run instructions
- Fix a few typos in other sections
- Add new sections for Troubleshooting and Training Resources
- Separate deployment info out into its own file, since the main readme is now getting quite long, intimidating and hard to navigate, and deployment is likely to be the last bit that new people will need

## Type of change

Documentation update; no changes to running code

# Checklist

- [] I have commented my code, particularly in hard-to-understand areas
- x[] My changes generate no new warnings

### DIFF
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -1,0 +1,99 @@
+This file gives instructions for deploying the API to AWS, and the app to TestFlight (iOS) and the Google Play Store (Android).
+
+- [Setup to deploy the app](#setup-to-deploy-the-app)
+- [API deployment on AWS](#api-deployment-on-aws)
+- [App deployment](#app-deployment)
+
+# Setup to deploy the app
+
+This is how you get set up ready to deploy the app to the Google Play Store (for Android) and TestFlight (for iOS) later, using [Fastlane](https://fastlane.tools/).
+
+You don't need to worry about doing this section until you're through all the steps above and you've solved any headaches getting the API and the app running locally.
+
+**If you're on a Mac** you can deploy the Android and iOS versions of the app.
+
+**If you're on Windows/Linux** you can only deploy the Android version of the app (you'll always need to get another team member with a Mac to deploy the iOS version).
+
+1. Install Fastlane: [Mac instructions in the 'Installing Fastlane' section here](https://thecodingmachine.github.io/react-native-boilerplate/docs/BetaBuild/#installing-fastlane) - [Windows/Linux instructions here](https://docs.fastlane.tools/getting-started/android/setup/)
+
+   ## Google Play Store (Android)
+
+2. Add the `key.json` file into the `/app/android/` directory. This file contains credentials for uploading the app to the Google Play Store. Ask on Slack for another developer in the team to send you it.
+
+   > Because this contains sensitive access credentials we should never commit this to GitHub as our repository is open-source, anyone can see it.
+
+3. Create a personal upload key to digitally 'sign' the app when you send it to the Google Play Store. Follow the instructions in the ['Generating an upload key' section of this page](https://reactnative.dev/docs/signed-apk-android#generating-a-signing-key). As part of doing that you'll be asked to set a password -- save this somewhere safe (e.g. [a password manager](https://www.techradar.com/uk/best/password-manager)), you'll need it in the future. Put the `my-release-key.keystore` file you create into the `/app/android` directory.
+   > This file also should never be committed to GitHub.
+
+## TestFlight (iOS)
+
+[To add]
+
+# API deployment on AWS
+
+June 2022:
+
+In the volapp-dev-test account, an [Elastic Beanstalk](https://eu-west-2.console.aws.amazon.com/elasticbeanstalk/home?region=eu-west-2#/environments) environment called Volunteerapp-env has been created (manually for now).
+
+You can connect your app to this environment by changing STA_BASE_URL to the load balancer address in `Volunteer-app/app/src/Config/index.ts`:
+
+` STA_BASE_URL: 'http://volunteerapp-env.eba-ivfm2tgp.eu-west-2.elasticbeanstalk.com',`
+
+Note - as we move this into IaC and set up some build pipelines, things like env names, app names, domain names, IP Addresses will probably change.
+
+For support, please @ David Calder in the [volunteer-app](https://scottishtecharmy.slack.com/archives/C01SUL6K5E1) Slack channel
+
+## Updating the API
+
+1. Git clone Scottish-Tech-Army/Volunteer-app to your computer
+2. `cd Volunteer-app/api`
+3. `zip ../myapp.zip -r * .[^.]*`
+4. Go to the AWS Management Console and navigate to Elastic Beanstalk.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -22,82 +23,108 @@ Ask Joanna on Slack to set you up, you'll get an email invitation to download Te
...
-7. Go to the `app` folder inside the project folder (e.g. **\Volunteer-app path\app**)
+6. At the command prompt type `npm install` then `npm start` to start the Volunteer App API server. You should see a message that says `Volunteer App API is listening on port <number>` - make a note of the port number for later
+
+7. Go to the `app` folder inside the project folder (e.g. **/path/to/Volunteer-app/app**)
 
 8. At the command prompt type `npm install`
-    >**Note:** Inside the `app` folder there is package-lock.json. Everytime this is modified, it is advised to repeat step 6 before  running the project.
 
-    >**Note:** you may need to run `npm install --legacy-peer-deps`
+   > **Note** Inside the `app` folder there is `package-lock.json`. Everytime this is modified, it is advised to repeat step 6 before running the project.
```
joefhall, line 77: Reading this again, wonder should this line say repeat this step (rather than step 6) -- and every time `package.json` or `package-lock.json` is modified?

And maybe we should say something similar above on step 6 itself for the `api` folder?

DEPLOYMENT.md
```diff
@@ -0,0 +1,100 @@
...
+
+- [Setup to deploy the app](#setup-to-deploy-the-app)
+- [API deployment on AWS](#api-deployment-on-aws)
+- [App deployment](#app-deployment)
+
+# Setup to deploy the app
+
+This is how you get set up ready to deploy the app to the Google Play Store (for Android) and TestFlight (for iOS) later, using [Fastlane](https://fastlane.tools/).
+
+You don't need to worry about doing this section until you're through all the steps above and you've solved any headaches getting the API and the app running locally.
```
joefhall, line 11: Tiny thing, but "all the steps above" should probably be something like "all the setup steps in the README"?

DEPLOYMENT.md
```diff
@@ -0,0 +1,100 @@
...
+
+8. Download the updated version of the app to your Android phone ([see download instructions](#download-the-app) near the top of this README). On the Google Play Store screen, there should be an 'Update' button to download the latest version of the app to your device.
+
+   ### Google Play Store known issues
+
+   - Near the end of the deployment process an AAB file is uploaded to the Google Play Store. This can take some time (e.g. 20 minutes on slow internet connections). It should be working, unless you get a `HTTPClient::SendTimeoutError: execution expired` error message in your terminal window. If the Fastlane process fails for this reason, you may need to run it again.
+
+   ## TestFlight (iOS)
+
+9. In `/app/src/Config/index.ts` set `STA_BASE_URL` to point to **the IP address** for the external URL for the API endpoint on AWS, see [known issues](#known-issues) above -- not to your localhost or its IP address.
```
joefhall, line 98: As this is covered in step 5 now (moved it to before we get to the Android/iOS specific as it applies to both) we can remove it from here now

joefhall (changes requested):
Thanks a lot for doing all this @rc-1988 -- this is excellent!  Just a few small suggestions

rc-1988:
Thanks @joefhall  - I think I can't see the wood for the trees any more, thank you for having a better eye than me! All addressed in c6906afa3e8ee6e8c57325669e4cf43531b6a590

README.md
```diff
@@ -29,94 +29,104 @@ Once that's done, you go to [this page](https://play.google.com/apps/internaltes
...
+9. If you are on a Mac and want to run the iOS build of the app, go to the `app/ios` folder in a terminal window. At the command prompt type `pod install`
 
-11. type in command terminal: `npm run ios` or `npm run android`
+10. Duplicate the example config file `app/src/Config/index.example.ts` and name your new file `app/src/Config/index.ts`
 
-# Subsequent run
+**Note** If the app has difficulty connecting to the API, you may need to specify your private IP address and API port number in `app/src/Config/index.ts`. The IP address settings may vary depending on the setup of your dev machine and local network. IP settings that have worked for others include: `127.0.0.1` (the loopback address); the private IP of the local dev machine; and the private IP of the default gateway for your network. You made a note of the API port number earlier. In the line `STA_BASE_URL: 'http://localhost:3000'`, replace `localhost` with your own IP, and if necessary, replace `3000` with the port number.
```
joefhall, line 87: Small extra thought on this: I've seen a couple of us (including me!) miss off the port number and/or inadvertently change the URL to `https` when faced with the headache of trying different options here.  I wonder if it would be useful to put in a full URL example with IP address to make it extra clear and help folks avoid that mistake, e.g. `http://127.0.0.1:3000`

joefhall (approved):
Great work @rc-1988 -- thanks again!

README.md
DougieWougie: Doesn't localhost always resolve to 127.0.0.1? Given that Android emulator treats that as the loopback device on the emulator, it might be clearer just to use http://127.0.0.1:3000?

rc-1988: I've tried to allow for different port numbers in case someone's machine decides to use something different for whatever reason.

There is an example a couple of lines down: "If your computer's private IP is 192.168.1.50 and the API is listening on port 3000, the line should now read STA_BASE_URL: 'http://192.168.1.50:3000'." - is there a better way to format it to make it stand out?

I think 127.0.0.1 was your suggestion @joefhall - has anyone had success with 127.0.0.1 but not with localhost? If not, then we could take that out.

joefhall: Possibly yes in my hazy memory back in the mists of time, but I'm not sure.  (I was basically thinking giving people options to try at this stage might be helpful -- only if they're actually useful of course! 🙂  -- as it can be such a frustrating experience.)

rc-1988: Thanks - that makes sense. Better to take it out and we can put it back in later; or leave it in on the off-chance it does work for anyone?

rc-1988: Better solution: [this](https://developer.android.com/studio/run/emulator-networking) from the Android dev docs suggests 10.0.2.2, and I can confirm that setting STA_BASE_URL to http://10.0.2.2:3000 works for me, so [b8e4ffe](https://github.com/Scottish-Tech-Army/Volunteer-app/pull/81/commits/b8e4ffeab6e6897bb5a5e7c94e514e490da25d95) puts that in instead of 127.0.0.1.

joefhall: I guess I would be inclined to include it because it gives a possible option to try.  Just did some looking around on Stack Overflow and I have also seen that `http://10.0.2.2:3000` works for some folks facing this same problem, so might be worth including that too.

joefhall: Sorry, ignore that last comment of mine! 😂   I hadn't refreshed the page and seen your latest one.  Good shout!  👍

DougieWougie (approved):
@rc-1988 - sorry thought I'd approved this already!

PR #114, opened 2023-02-10
SVA-260 update bottom nav bar with STA icons and NativeBase

PammyLo:
- [Description](#description)
- [Checklist](#checklist)

# Description

- The new STA icons for the bottom navigation bar have been added as part of a custom library with only selected icons we used
- Bottom navigation bar now uses these icons 
- Design has been updated to use Native Base
- Colours work as design spec for light and dark mode 

## Type of change

# Checklist


PammyLo:
Most of the ES Lint errors and warnings have been addressed but it’s hard to check as it doesn’t seem to update when I re-run it and the latest commits don’t seem to have been taken into account. Have even pushed an almost empty commit to see if it would trigger a proper re-run but to no avail. Will try again later. Perhaps @rc-1988 can help resolve that? 

In the meantime would be most obliged to get reviews, thanks!

app/src/Navigators/Main.tsx
```diff
@@ -1,62 +1,144 @@
...
-        name="Projects"
-        component={ListContainer}
-        initialParams={{ type: ListType.Projects }}
+        name="Example Container"
+        component={VerticalStackContainer}
         options={{
-          headerShown: false,
-          tabBarIcon: () => (
-            <Entypo color={Colors.text} name="home" size={24} />
+          tabBarIcon: ({ focused }) => (
```
joefhall, line 110: We're getting a React warning message here "Do not define components during render."  Would be good to address this (and for the other `tabBarIcon`s).

I think what you've got here is nearly right, but instead of calling a component (like `HomeTabBarIcon`), call a function which renders a component -- a bit like [this one.](https://github.com/Scottish-Tech-Army/Volunteer-app/commit/722b209d0e5d3bf5719770b492cfe9df3be5c683#diff-08c2d3e3529a4185692086d551e348cf1ea0f234b50783e269e3372dc364547dR33)

Also, as the current `HomeTabBarIcon`, `ProjectsTabBarIcon`, etc mostly share the same code, would be good if a single render function could replace all of them and you could just pass in as parameters like focused, colorMode and icon (or a variable that can be used to determine which icon to use).

Give me a shout if that doesn't make sense or you need any help.

app/src/Navigators/Main.tsx
```diff
@@ -1,62 +1,144 @@
...
+    ? StaTheme.colors.text['100']
+    : StaTheme.colors.textDarkMode['100']
+}
+
+const HomeTabBarIcon = ({ focused, colorMode }: navBarProps) => (
+  <>
+    <TabBarIconWrapper focused={focused} colorMode={colorMode} />
+    <View
+      style={{
+        padding: 4,
```
joefhall, line 54: How come the padding varies for each icon -- is it because the icons are different sizes?  If so, there might be another way to approach this, with a layout that centres the icon inside an area?

app/src/NativeBase/Assets/Icons/index.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * @file Exports icons for easy access elsewhere.
+ */
+
+export { default as Home } from './Home.svg'
+export { default as Projects } from './Projects.svg'
+export { default as EventTicket } from './EventTicket.svg'
```
joefhall, line 7: For consistency, if we're naming icons here by what they are for (e.g. Projects) rather than what they look like visually, I think it would be more consistent to just call this `Events`

joefhall (changes requested):
Nice work @PammyLo!  Have added a few questions/suggestions

app/src/NativeBase/Assets/Icons/index.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * @file Exports icons for easy access elsewhere.
+ */
+
+export { default as Home } from './Home.svg'
```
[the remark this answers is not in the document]
PammyLo: I thought they were but worth checking!

app/src/Navigators/Main.tsx
PammyLo: Thanks @joefhall, all valid points, now resolved :)

app/src/NativeBase/Assets/Icons/index.ts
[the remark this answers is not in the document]
PammyLo: I checked with Mia and the icons are from iconify. Instead of adding more libraries for each new icon I have started to collect them in a custom library containing a selection of icons we use, starting with these four. Over time we can add the other icons we use to this library which will reduce the number of load time once the design is more final and the app goes into prod. 
The previously added SVGs have been deleted. 
The app may need npm install again.

app/src/NativeBase/Assets/Icons/index.ts
PammyLo: You're quite right Joe, this should be more consistent. I've given it some thought and very much prefer icon names that describe what they look like, making it easier to use them for other relevant purposes if appropriate. Some icons of course are an exception to that rule (such as home and search), and in this case the home icon is still called home.

app/src/Navigators/Main.tsx
PammyLo: Oops, think it's you that's supposed to click resolved? ;)

app/src/NativeBase/Assets/Icons/selection.json
```diff
@@ -0,0 +1,126 @@
+{
```
joefhall, line 1: Might be good to name this file something like `icomoon_selection.json` so it's even clearer what it's for (or create a separate sub-directory for the Icomoon files so they can sit together)?

app/src/Navigators/Main.tsx
```diff
@@ -1,63 +1,123 @@
...
+    <View
+      borderTopColor={
+        focused
+          ? colorMode === 'light'
+            ? StaTheme.colors.primary['100']
+            : StaTheme.colors.primary['40']
+          : 'transparent'
+      }
+      borderTopWidth={2}
+      style={{
```
joefhall, line 58: It looks like the recommended way to do [width, height, etc](https://docs.nativebase.io/utility-props#h3-layout-width-and-height) and [flexbox properties](https://docs.nativebase.io/utility-props#h3-flexbox) is to use named utility props (`width=` etc) rather than one `style=` object?

app/src/Navigators/Main.tsx
```diff
@@ -1,63 +1,123 @@
...
+    case 'Events':
+      iconName = 'ticket'
+      break
+    case 'Profile':
+      iconName = 'user'
+      break
+  }
+
+  return (
+    <View
```
joefhall, line 49: On my Android emulator the text below the icons looks very close to the bottom of the screen -- might need some extra bottom padding in order to make this more comfortable and match the Figma designs?

Screenshot 2023-02-20 18 29 51 - Display 2

joefhall (changes requested):
Great stuff, thanks @PammyLo!  I just spotted a couple of other things.

PammyLo:
Thanks again Joe, that does looks better with a bit of margin, have also improved the colours and renamed some of the components for clarity, and moved the icomoon stuff into a subfolder. 
:)

app/src/Navigators/Main.tsx
PammyLo: Thanks Joe, you're right. These were the default position of the tab labels. The inactive colour is also a lighter default. Looking better at our design it is indeed different, have added custom label styling now and removed the vague grey line at the top of the nav bar.

joefhall (approved):
Nice work on all of this @PammyLo -- thanks a lot!  Great stuff 🙌

g-dunlop (approved):
This looks great @PammyLo, nice one!  Some of the code is over my head just reading through if I'm being honest, but it runs perfectly for me and looks fantastic

PR #118, opened 2023-02-27
:pencil: OPENSSL error fix added to README

treneff:
- [Description](#description)

# Description

Updated README with fixes regarding OPENSSL and Node 17+ conflicts on bundling of the application


DougieWougie (commented):
Hi @treneff I'd rephrase that a bit because the line above says to use Node 16 (which was LTS when the Boilerplate was adopted).

Recommending nvm/nvs should probably have its own point (it does on the graphical guides in Whimsical).

DougieWougie (approved):
Thanks @treneff!

joefhall (approved):
Nice one, thanks @treneff!

PR #119, opened 2023-03-04
Update README.md

vglampard:
Added Graeme's fix for the 'normalize-color' issue I was getting into the troubleshooting section.


joefhall:
Hi @vglampard please can you delete this pull request now as it's no longer needed (Graeme merged another PR into `main` which includes `@react-native/normalize-color` so the problem you ran into should now be automatically fixed for anyone in the future)

PR #124, opened 2023-03-10
SVA-275

vglampard:
- [Description](#description)
- [Checklist](#checklist)

# Description

- The issue is that we wanted to add the ability to version the API going forward.
- So I have hard coded 'v1' onto the end of the existing "/" endpoint in app.js.  
- Then, I have added a STA_API_VERSION variable of "/v1" to the config object in config.js. In order for this change to actually be pushed up, I've also added this additional variable to config.example.ts 
- I have then added this after any instance of STA_BASE_URL in the code to make sure each request is hitting the right /v1 API endpoint. 
- This will cause problems, as in order to be able to access the STA_API_VERSION variable, and to update the config alias, people will still need to change their actual local config.ts file! Is this OK?

Fixes # SVA-275

## Type of change

This will break API fetches for people if they don't add the new variable in their actual config file?

# Testing locally

- Need to add the new STA_APP_VERSION variable in config.example.ts to your actual config.ts file. 
- I am going to see what needs to be fixed re: linting when it goes through the CI testing just now

# Checklist

- [] My changes generate no new warnings

### DIFF
--- a/api/app.js
+++ b/api/app.js
@@ -5,7 +5,7 @@ const cors = require('cors');
 
 app.use(cors());
 app.use(express.json());
-app.use('/', routes);
+app.use('/v1', routes);
 
 app.use((req, res, next) => {
   const err = new Error("Something went wrong. Please try again.");
--- a/app/src/Config/index.example.ts
+++ b/app/src/Config/index.example.ts
@@ -7,4 +7,6 @@ export const Config = {
   // STA Base API URL (localhost may not work on Android as it points to
   // the local device, you can try using your local IP address instead)
   STA_BASE_URL: 'http://localhost:5000',
+  // Prefix that specifies current API version.
+  STA_API_VERSION: '/v1',
 }
--- a/app/src/Services/modules/events/fetchAll.ts
+++ b/app/src/Services/modules/events/fetchAll.ts
@@ -5,6 +5,7 @@ import { EventsRange } from '.'
 
 export default (build: EndpointBuilder<any, any, any>, range: EventsRange) =>
   build.query<Events, string>({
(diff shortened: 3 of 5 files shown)


joefhall (changes requested):
Looking good @vglampard, thanks!

Please could you also amend `api/definitions/api_definition.json`  [It looks like](https://stackoverflow.com/a/73698584/15422951) this should be added between the current lines 7 and 8:

```
"servers": [
    {
      "url": "/v1"
    }
  ],
```

joefhall (approved):
Nice one, thanks @vglampard ! 🙌

vglampard:
@DougieWougie @M4RC1N76 looks like everything's OK to go on this one too!

vglampard:
OK @DougieWougie @M4RC1N76 is this one also finally good to kill off?

DougieWougie (approved):
Thanks for sticking with it @vglampard !

joefhall:
> Just to DOUBLE CHECK! I can go ahead and merge this, and put a note in slack that people will need to update their local config.ts file? @DougieWougie @joefhall

Hi @vglampard yep exactly, and tell folks that this will only work with their local running API until we've deployed the API to the server.

We'll need to deploy the API to the server and make sure that happens at the same time as the app is next deployed, so they're in sync.

DougieWougie:
I've containerised the API @joefhall and was planning to automate it's deployment with GitHub Actions. I'll wait out for now!

joefhall:
> I've containerised the API @joefhall and was planning to automate it's deployment with GitHub Actions. I'll wait out for now!

That would be great @DougieWougie!  Don't want to hold you back, but maybe we just deploy this change first.

PR #121, opened 2023-03-06
SVA-250 | Get Vimeo, YouTube and other project videos working

joefhall:
- [Description](#description)
  - [Type of change](#type-of-change)
- [Testing locally](#testing-locally)
- [Checklist](#checklist)

videos

# Description

## Problems
- Most project videos were no longer playing in the app
- In some cases there was a settings/permissions issue, now fixed with Joanna updating settings in Vimeo
- But the Vimeo public API was still not providing MP4 URLs for most of our videos -- turns out they're deprecating this (unofficial) service, so we can't get direct MP4 URLs any more and we need to change how we load videos in the app
- Also, the app already didn't play YouTube videos -- we had assumed charities would always provide a video on Vimeo, but sometimes they've been supplying a YouTube URL so we need to cater for that
- It also turns out there are edge cases where charities occasionally provide different video links entirely, e.g. one recent project linked to video page on ITV news site -- it would be good to be able to show these too

## Solutions
- API code gets a "player only" URL for Vimeo and YouTube videos, which is a very simple web page that contains only the actual video player itself
- In the app we show this "player only" URL in a webview (embedded web browser) in the app (from reading discussions on Stack Overflow this seems to be how others are playing in-app videos hosted on these sites; there is a YouTube player npm package but it uses an insecure, deprecated API so is best avoided)
- Inspired by good point made by @PammyLo, added some error handling to the Video component too
- Also added some JSDocs I forgot on previous PRs (sorry!)

Fixes SVA-250

## Type of change

Please delete options that are not relevant

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New Feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# Testing locally

- You need to run `npm install` (or `npm install --legacy-peer-deps`) inside your `app` directory
- **Mac users:** you'll also need to run `pod install`.  This should update `app/ios/Podfile.lock` Please can you commit and push this to this branch so it's included in this pull request (if any other iOS files are changed please include them too).
- If you want to try running the cron job to update the projects AirTable database table with latest data from Jira and latest video URLs, you will need some extra variables added to your `.env` file -- DM me on Slack for these.  Then in your terminal go to the `api` directory then run the command `node cron-jobs/run-projects.js`
- 'MADE 4 U IN ML2' is one of the projects that will show `video_webpage_player_only` in a webview
[...]

PR #127, opened 2023-03-17
Run cron jobs from server.

DougieWougie:
- [Description](#description)
- [Checklist](#checklist)

# Description

- Cron jobs
- Added Dockerfile and .dockerignore

Fixes # SVA-224

## Type of change

# Testing locally

- Is it this simple?
- I've created Docker container so that if it isn't then I just add to crontab plus Docker is easier than faffing about zipping files and can be automated through GitHub Actions.

# Checklist

- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,10 +133,6 @@ Below are some commonly encountered issues and possible ways to resolve them. If
 
    > **On Android,** if you get an error message that includes `INSTALL_FAILED_UPDATE_INCOMPATIBLE` this may be because you previously installed a newer version of the app for your emulator (e.g. on a new branch or testing someone else's pull request) then you switched back to an earlier version. Uninstall the app from your emulator with the command `adb shell pm uninstall org.scottishtecharmy.volunteerapp` then run `npm run android` again.
 
-4. Optional: Add event video thumbnail images and update event video files whose URLs expire after an hour. Open another command terminal window, go to the `api` folder inside the project.
-   - If you want to manually add event video thumbnails, enter this command: `node cron-jobs/run-events.js`
-   - See the previous step above on how to run all cron jobs automatically.
-
 # Development
 
 ## App
--- a/README.md
+++ b/README.md
@@ -133,15 +133,7 @@ Below are some commonly encountered issues and possible ways to resolve them. If
 
    > **On Android,** if you get an error message that includes `INSTALL_FAILED_UPDATE_INCOMPATIBLE` this may be because you previously installed a newer version of the app for your emulator (e.g. on a new branch or testing someone else's pull request) then you switched back to an earlier version. Uninstall the app from your emulator with the command `adb shell pm uninstall org.scottishtecharmy.volunteerapp` then run `npm run android` again.
 
-4. Optional: Update the cached projects data from Jira _(during development, you probably only need to use this if you need the very latest data from Jira or to update project video files whose URLs expire after an hour)_. Open another command terminal window, go to the `api` folder inside the project.
-
-   - If you want to manually update cached projects data, enter this command: `node cron-jobs/run-projects.js`
-     > During development, it's preferable to do this than to run the scheduled cron job described below.
-   - If you want to automatically run all cron jobs regularly , enter this command instead: `node cron-jobs/run-cron-jobs.js` Leave this terminal window open as long as you want this to keep running.
-     > See more about [cron jobs](#cron-jobs) below
-     > Be careful if using this during development: if multiple developers are running this simultaneously, these could conflict if more than one person is updating the same AirTable tables at the same time.
-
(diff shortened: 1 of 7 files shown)


api/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+WORKDIR /usr/src/api
+
+# Install app dependencies
+# A wildcard is used to ensure both package.json AND package-lock.json are copied
+# where available (npm@5+)
+COPY package*.json ./
+
+RUN npm install
+# If you are building your code for production
+# RUN npm ci --only=production
```
joefhall, line 13: Should this be uncommented (as we're basically using for this for production from now on with the app launching soon)?

api/server.js
```diff
@@ -1,5 +1,7 @@
 const app = require("./app");
+const cron = require('./cron-jobs/run-cron-jobs')
```
joefhall, line 2: Tiny thing, but this and line 5 should have a semi-colon at the end.

(At some point we should switch the API to Typescript and could enforce the same coding style as the front-end app e.g. no semi-colons for line endings)

joefhall (changes requested):
Thanks for picking this up @DougieWougie!  Be great to have this up and running.

Just had a question and spotted a tiny code formatting thing.

api/Dockerfile
DougieWougie: I thought that but it didn't work, something to do with Cors. I'll have another look.

joefhall:
P.S.  When I tried running it locally the projects cron job failed, but this will be fixed by [this PR](https://github.com/Scottish-Tech-Army/Volunteer-app/pull/121)

api/Dockerfile
joefhall: Cool no worries, I just wondered about it based on the comment.  (Or if we don't need it could remove those two lines?)

DougieWougie: Might be cors is a Dev dependency when it shouldn't be? I'll check.

DougieWougie: It was indeed cors installed as devDependencies.

joefhall:
Hi @DougieWougie as this change will mean people running the API locally and now automatically running cron jobs, I think we need to update the README too if you could include that in this PR.  Basically point 4. of 'Subsequent run' (lines 136-143) can be removed now as it'll no longer apply after this change has been made (and point 5. below it needs to be renumbered to 4.)

Also we'll need to get everyone to change their .env file to point to a test projects AirTable table (we've got one set up already) when developing locally, rather than the table the app will use in production (otherwise there could be a load of people with cron jobs all updating the production projects table which could be risky).  Could you update DEPLOYMENT.md as well -- in the 'Updating the API' section I think we could add an extra step between the current 2 and 3 asking people to set `AIRTABLE_PROJECTS_RESOURCES_CACHE_TABLE` to the name of the production table (instead of the test table) in their api/.env file.

Thanks a lot

README.md
```diff
@@ -133,15 +133,7 @@ Below are some commonly encountered issues and possible ways to resolve them. If
...
-
-   - If you want to manually update cached projects data, enter this command: `node cron-jobs/run-projects.js`
-     > During development, it's preferable to do this than to run the scheduled cron job described below.
-   - If you want to automatically run all cron jobs regularly , enter this command instead: `node cron-jobs/run-cron-jobs.js` Leave this terminal window open as long as you want this to keep running.
-     > See more about [cron jobs](#cron-jobs) below
-     > Be careful if using this during development: if multiple developers are running this simultaneously, these could conflict if more than one person is updating the same AirTable tables at the same time.
-
-5. Optional: Add event video thumbnail images and update event video files whose URLs expire after an hour. Open another command terminal window, go to the `api` folder inside the project.
+4. Optional: Add event video thumbnail images and update event video files whose URLs expire after an hour. Open another command terminal window, go to the `api` folder inside the project.
```
joefhall, line 136: run-cron-jobs.js actually runs both projects and events, so suggest that either remove 4. here and the two bullet points that follow (lines 137-8) -- or in api/server.js require run-projects rather than run-cron-jobs (that way we'd only be running projects cron jobs and you can leave the README here as you've got it now).

The former sets us up ready for events in the future, but I guess the latter (projects only) is all we strictly need for MVP.

api/server.js
joefhall, line 2: Tiny thing, but semi-colon still needed here 🙂

joefhall (changes requested):
Thanks for the changes, just spotted a couple of small things 🙂

api/server.js
DougieWougie: Done. Really this time.

M4RC1N76 (approved):
I have familiarise myself with changes in the files. Great work @DougieWougie !!

PR #131, opened 2023-03-27
SVA-258/NativeBaseProjectList

CatMac228:
- [Description](#description)
- [Checklist](#checklist)

# Description

- Updated List container and associated components to match the new Project list design using Native Base.
- Moved the container and associated components to the Native Base folder.
- Updated imports on other components to point to the new Native Base folder.

Fixes # [SVA-258](https://sta2020.atlassian.net/browse/SVA-258)

## Type of change

# Testing locally

- No additional requirements to test these changes locally.
- The 'buddying' icon is different from the design as was unable to import the one used in Figma from Iconify. Have added a temporary one while we try to figure it out :) 

# Checklist

IOS: 

Android:


DougieWougie (changes requested):
Great work - looks so much more like an App now!

app/src/NativeBase/Components/List.tsx
```diff
@@ -4,13 +4,27 @@
...
+  Pressable,
+  ShareIcon,
+  Text,
+  VStack,
+  Card,
+  Icon,
+} from 'native-base'
 
 // General styled components
 const NoneFound = styled.Text`
```
joefhall, line 30: As `NoneFound` and `TryAnotherSearch` are used by projects, please can you also convert these to using NativeBase instead of `styled`.  (No worries about the events-specific things still using `styled`)

app/src/NativeBase/Components/List.tsx
```diff
@@ -116,18 +130,65 @@ const List: FC<ListProps> = ({ data, mode, options, searchScreen, type }) => {
...
           renderItem={({ item }) => {
             return (
-              <ProjectListItem
-                onPress={() => {
-                  navigate('ProjectDetail', { item, key: item.res_id })
-                }}
-              >
-                <ProjectSummary project={item} />
-              </ProjectListItem>
+              <Card margin="2">
```
joefhall, line 135: From looking at Figma, the edge of the card should line up horizontally with the edge of the segmented picker above, whereas the card has extra spacing around it at the moment here.  So I think the margin setting here should be changed to `marginY="2"` (i.e. vertical margin only, not horizontal) and then I think it'll line up (though do check).

app/src/NativeBase/Components/List.tsx
```diff
@@ -116,18 +130,65 @@ const List: FC<ListProps> = ({ data, mode, options, searchScreen, type }) => {
...
+                        onPress={underDevelopmentAlert}
+                      />
+                      <FavouriteIcon
+                        size="md"
+                        color="accentGrey.100"
+                        onPress={underDevelopmentAlert}
+                      />
+                    </HStack>
+
+                    <Text fontSize="xs">{item.client}</Text>
```
joefhall, line 163: It looks like there's much bigger vertical spacing above especially, but also below, the organisation name than on the Figma designs?  Not sure if it's to do with how the `VStack` spacing is working and/or some extra space inside the `HStack` above?

app/src/NativeBase/Components/List.tsx
```diff
@@ -116,18 +130,65 @@ const List: FC<ListProps> = ({ data, mode, options, searchScreen, type }) => {
...
+                    <ColouredTag title={item.role} />
+                    {Boolean(item.buddying) && (
+                      <>
```
joefhall, line 166: Small thing but you don't need the `<>` and `</>` here, as the `HStack` works as a top-level element

app/src/NativeBase/Components/List.tsx
joefhall, line 135: I'm finding it slightly hard to work out from the Figma design how much vertical padding there should be inside the card, but feels like this maybe needs `paddingY="2"`?

app/src/NativeBase/Components/List.tsx
```diff
@@ -116,18 +130,65 @@ const List: FC<ListProps> = ({ data, mode, options, searchScreen, type }) => {
...
+                        <HStack
```
joefhall, line 167: On the Figma designs 'Suitable for pairing' goes below the description?

app/src/NativeBase/Components/Project/ProjectRoleTitle.tsx
```diff
@@ -0,0 +1,38 @@
+/**
```
joefhall, line 1: I think this component isn't being used anywhere and can be deleted?

app/src/NativeBase/Containers/ListContainer.tsx
```diff
@@ -7,23 +7,30 @@
...
 /* eslint-disable @typescript-eslint/no-shadow */
 
+import { Text, Heading, VStack } from 'native-base'
+
 import React, { useEffect, useState } from 'react'
-import { Text } from 'react-native'
 import { useDispatch, useSelector } from 'react-redux'
 import styled from 'styled-components/native'
-import { EventSearch } from './EventSearchContainer'
+import { EventSearch } from '../../Containers/EventSearchContainer'
```
joefhall, line 15: Small thing but the syntax we normally use to reference a different top-level directory inside `app/src` is '@/Containers/EventSearchContainer'

app/src/NativeBase/Containers/ListContainer.tsx
```diff
@@ -36,25 +43,16 @@ import {
 } from '@/Services/modules/projects'
 import { EventsState, setEvents } from '@/Store/Events'
 import { ProjectsState, setProjects } from '@/Store/Projects'
-import Theme from '@/Theme/OldTheme'
+import underDevelopmentAlert from '@/Utils/UnderDevelopmentAlert'
 
 const ClearSearchLabel = styled.Text`
```
joefhall, line 48: Please can you remove these `styled` components entirely and replace them with NativeBase components e.g. Text where they're used below

app/src/NativeBase/Containers/ListContainer.tsx
```diff
@@ -261,12 +267,21 @@ const ListContainer = (props: {
   }, [params?.options?.events, params?.search, params?.type])
 
   return (
-    <Theme>
-      <SafeArea>
-        <TopOfApp
-          showSearchButton
-          onSearchButtonPress={() => navigate(screens.search[params.type], '')}
-        />
+    <>
```
joefhall, line 270: I think instead of `<>`...`</>` things should be wrapped inside a NativeBase `ScrollView` here https://docs.nativebase.io/scrollview (like in e.g. ProfileContainer)

app/src/NativeBase/Containers/ListContainer.tsx
```diff
@@ -261,12 +267,21 @@ const ListContainer = (props: {
...
+      <TopOfApp
+        showSearchButton
+        onSearchButtonPress={() => navigate(screens.search[params.type], '')}
+      />
+      <VStack
+        _dark={{ backgroundColor: 'bgDarkMode.100' }}
```
joefhall, line 276: I think you can remove these `_dark` and `_light` settings here as the default does the same?

app/src/NativeBase/Containers/ListContainer.tsx
```diff
@@ -261,12 +267,21 @@ const ListContainer = (props: {
...
+        _light={{ backgroundColor: 'bg.secondary' }}
+        paddingBottom="2"
+        alignItems="center"
+        space={4}
+        padding={4}
+      >
+        <Heading fontSize="md">Projects List</Heading>
```
joefhall, line 283: From looking at [the NativeBase docs for Heading](https://docs.nativebase.io/heading), I think the recommended way to change a heading is using the `size` attribute rather than `fontSize` (works differently to e.g. `<Text>`

joefhall (changes requested):
Nice work @CatMac228!  Have spotted some things, but they're all basically small things so hopefully not loads to do

app/src/NativeBase/Components/List.tsx
joefhall: Actually, don’t worry about changing these — as I’m about to replace how the search results work in SVA-264

app/src/NativeBase/Containers/ListContainer.tsx
joefhall: Ignore this too — I’m going to replace how search results work right after you’ve done this ticket

app/src/NativeBase/Containers/ListContainer.tsx
```diff
@@ -36,25 +43,16 @@ import {
 } from '@/Services/modules/projects'
 import { EventsState, setEvents } from '@/Store/Events'
 import { ProjectsState, setProjects } from '@/Store/Projects'
```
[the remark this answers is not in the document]
CatMac228: It looks like their being used to store data in the Redux store - _"When data has loaded, store data in the Redux store so it can be used by other containers/components"_

Will be honest - have no idea what that means  @joefhall I think you might have added the code for this?

app/src/NativeBase/Containers/ListContainer.tsx
```diff
@@ -113,6 +111,14 @@ const ListContainer = (props: {
     } as Screens,
   }
 
+  const projectListOptions = ['all', 'saved', 'my'].map(
```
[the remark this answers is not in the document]
CatMac228: This was the red squigglys I was referring to on Friday - I could see them but Pam couldn't 😄

Is this syntax correct? 

const screens = {
    list: {
      [ListType.Events]: 'Events',
      [ListType.Projects]: 'Projects' as keyof RootStackParamList,
    } as Screens,

app/src/NativeBase/Containers/ListContainer.tsx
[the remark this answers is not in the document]
joefhall: Yep they’re being used further down in this file — when the container has loaded projects or events data, it stores it in the Redux store for use by other components e.g. projects search.

app/src/NativeBase/Containers/ListContainer.tsx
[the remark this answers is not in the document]
joefhall: Yep perfect, thanks 👍 (If you could do the same on the Events line while you’re at it, that’d be great.) The squigglies should then disappear from there, but give me a shout if they don’t 🙂

app/src/NativeBase/Containers/ListContainer.tsx
CatMac228: Getting this error if I add scrollView: 

`VirtualizedLists should never be nested inside plain ScrollViews with the same orientation because it can break windowing and other functionality - use another VirtualizedList-backed container instead. `

Seems like it doesn't like `<List>` sitting inside it. Maybe it needs to be something like a sectionList from native base instead? 🤔 https://docs.nativebase.io/section-list

app/src/NativeBase/Components/List.tsx
CatMac228: Did not know about X & Y margins! The spacing was driving me mad 😄

app/src/NativeBase/Components/List.tsx
CatMac228: I had padding in the VStack - looks better with it removed although there's one project which still has more spacing than the others...which is odd...

Screenshot 2023-04-17 at 16 56 04

Screenshot 2023-04-17 at 16 56 21

app/src/NativeBase/Containers/ListContainer.tsx
joefhall: Aha, no actually I think the problem is - I'm wrong! 🙂  I think you can just leave this as is.

app/src/NativeBase/Components/List.tsx
joefhall: Looking a lot better, thanks 👍   Might be good enough for now, unless you can easily figure out what's going on with that errant one.

app/src/NativeBase/Components/List.tsx
```diff
@@ -116,18 +130,62 @@ const List: FC<ListProps> = ({ data, mode, options, searchScreen, type }) => {
       return (
         <FlatList
```
joefhall, line 131: The one last thing I've spotted is that the bottom of the last item in the list is being cut off.  

I think (but am not certain) that you might need to set a `maxHeight` for the `FlatList` using an approach similar to what's being done in [ProjectRegisterInterest](https://github.com/Scottish-Tech-Army/Volunteer-app/blob/main/app/src/NativeBase/Components/Project/ProjectRegisterInterest.tsx) -- define a height for the stuff that's above the list (like line 37), calculate the height of the window (line 158) , then set a `maxHeight` of one minus the other (like line 171 except that uses `minHeight` whereas I think you want to use `maxHeight` here).

joefhall (changes requested):
Looking great @CatMac228 -- just one thing I spotted with the bottom of the last card being cut off when I see it in on my Android emulator.  Have put in a suggestion you could try, if that doesn't work or you can't figure it out and need some help, do give me a shout if you want.

app/src/NativeBase/Components/List.tsx
CatMac228: thank you for spotting, that's an embarrassing miss on my part! 😄 messed about with it there and it is kind of working but there's a bit of white space above the nav icons now...not sure if that's okay or not 

(red line = empty space)

Screenshot 2023-04-18 at 20 57 27

joefhall: @CatMac228 that's looking cut off on my emulator?

Screenshot 2023-04-19 10 26 14

CatMac228: oh no it looks so different on IOS 😢
Screenshot 2023-04-19 at 12 15 03

CatMac228: they seem to be behaving the same way when I use minHeight but not maxHeight 🤔

joefhall: It's possible I'm wrong about `maxHeight` and it should be `minHeight` -- was just a suggestion off the top of my head.  If you've got a version that works do push again and I can try it my end too.  I'm travelling this afternoon but should be able to get online again this evening.

CatMac228: No probs! Will remove the maxHeight bit for now although the list will be cut off slightly at the bottom. Wonder if it could be related to these issues https://github.com/facebook/react-native/issues/23693 &https://github.com/facebook/react-native/issues/36447 ?

joefhall (approved):
🙌

PR #133, opened 2023-03-31
SVA-281 | Hide welcome screen after first time the user sees it.

kelsiesmurphy:
- [Description](#description)
- [Checklist](#checklist)

# Description
### Problem
As a user, I only want to see the app welcome screen the first time I use the app.

### Solutions
Needs an update to the WelcomeButton handlePress function, to call the Redux dispatch method like in the ProfileContainer onChangeSplash function (and will need to access welcomeState using useSelector like in ProfileContainer).

Fixes # SVA-281

## Type of change

- [ ] This change requires a documentation update

# Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have addressed accessibility, if needed
- [ ] My changes generate no new warnings

### DIFF
--- a/app/ios/Podfile.lock
+++ b/app/ios/Podfile.lock
@@ -269,18 +269,18 @@ PODS:
     - React-jsi (= 0.65.1)
     - React-perflogger (= 0.65.1)
   - React-jsinspector (0.65.1)
-  - react-native-date-picker (4.2.9):
+  - react-native-date-picker (4.2.2):
     - React-Core
   - react-native-flipper (0.113.0):
     - React-Core
-  - react-native-safe-area-context (3.4.1):
+  - react-native-safe-area-context (3.3.2):
     - React-Core
-  - react-native-video (5.2.1):
+  - react-native-video (5.2.0):
     - React-Core
-    - react-native-video/Video (= 5.2.1)
-  - react-native-video/Video (5.2.1):
+    - react-native-video/Video (= 5.2.0)
+  - react-native-video/Video (5.2.0):
     - React-Core
-  - react-native-webview (11.26.1):
+  - react-native-webview (11.26.0):
     - React-Core
   - React-perflogger (0.65.1)
   - React-RCTActionSheet (0.65.1):
@@ -348,13 +348,13 @@ PODS:
     - React-perflogger (= 0.65.1)
   - RNCalendarEvents (2.2.0):
     - React
-  - RNCAsyncStorage (1.17.11):
+  - RNCAsyncStorage (1.17.7):
     - React-Core
   - RNGestureHandler (1.10.3):
     - React-Core
   - RNPDF (0.14.0):
     - React
-  - RNScreens (3.20.0):
+  - RNScreens (3.14.0):
     - React-Core
     - React-RCTImage
   - RNSVG (12.5.1):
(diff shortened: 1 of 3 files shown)


joefhall:
Nice one, thanks @kelsiesmurphy!  I've just spotted this is failing the ES Lint checks -- not because of anything you've done 🙂  but because of some formatting issues with the existing code (I think the code in that file was added before we set up the automated checks).

You'll need to fix those or GitHub won't let you merge this in -- if you click on 'Details' below next to the check that's failing, it should tell you what needs changing.

kelsiesmurphy:
Oops, thanks for that Joe, they're now passing. Let me know if you need more detail for the JSDocs side of things.

app/ios/Podfile.lock
```diff
@@ -269,18 +269,18 @@ PODS:
     - React-jsi (= 0.65.1)
```
joefhall, line 269: As this PR isn't about making any package changes, I don't think you should need to include this file?

app/ios/STA.xcodeproj/project.pbxproj
```diff
@@ -994,7 +994,7 @@
 				COPY_PHASE_STRIP = NO;
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
 				ENABLE_TESTABILITY = YES;
-				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = "arm64 ";
+				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = "";
```
joefhall, line 997: Also not sure these changes should be included in this PR.  If they're general changes that do need to be made for the iOS setup, I think that should go in a separate PR (and so it can be flagged for testing by other Mac users, etc).

app/src/Components/Welcome/WelcomeButton.tsx
```diff
@@ -1,15 +1,23 @@
-//'Get Started' button allows user to go to main part of app from any slide
-
+/**
+ * @file 'Get Started' button allows user to go to main part of app from any slide
+ */
 import React from 'react'
-import {TouchableOpacity, Text} from 'react-native'
+// import { TouchableOpacity, Text } from 'react-native'
```
joefhall, line 5: Please can you remove this line if no longer needed, rather than commenting it out

joefhall (changes requested):
Nice work, thanks @kelsiesmurphy!  I spotted a few little things which have crept in, if you could take a look -- but the functionality change itself looks good to me 👍

app/src/Components/Welcome/WelcomeButton.tsx
kelsiesmurphy: I'm not sure what this is actually, both were flagged in the linting checks for going through unused, but I'll remove them for now.

app/ios/Podfile.lock
kelsiesmurphy: This is no longer included.

app/src/Components/Welcome/WelcomeButton.tsx
kelsiesmurphy: This line has been removed

joefhall (approved):
Good work @kelsiesmurphy -- thanks for all of this! 🙌

PammyLo (approved):
works fine on iOS too :)
Great work @kelsiesmurphy, thanks

PR #155, opened 2023-04-28
Adding the colour changes necessary

kelsiesmurphy:
- [Description](#description)
- [Checklist](#checklist)

# Description

Changes reference image

### To be done:
The dark mode was asked to be adjusted, to do the following:
- Update the dark mode background colour to the hex code: #373735
- Ensure the card colour remains the same as before despite the above change (hex code should be: #3c3c3b)
- Remove the light grey border currently visible in the dark mode (ensuring the light mode remains the same.)

### My approach
- Added a new common colour option to the theme (darkGrey200), which is the #373735 hex code. I was then able to update the places where the bgDarkMode is referenced across multiple files by changing them from the 100 variant to the 200 variant. This also ensured that any existing references to the 100 variant across the app remained consistent. Happy to update this is this does not fit the current NativeBase best practices.

Note: The All/Saved/My tab section at the top of the project-list screen has not been updated to match the Figma design in this ticket. If it is required please let me know, but I understood it as being out of scope for this ticket.

Fixes # SVA-298

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# Testing locally

- Run the app in an emulator and check that light mode has not changed anywhere, then check the dark mode to ensure it matches the design reference above.

# Checklist

### DIFF
--- a/app/src/NativeBase/Theme/StaTheme.tsx
+++ b/app/src/NativeBase/Theme/StaTheme.tsx
@@ -67,8 +67,7 @@ const StaTheme = extendTheme({
       100: common.colours.offWhite100,
     },
     bg: {
-      100: common.colours.white,
-      secondary: common.colours.offWhite100,
+      100: common.colours.offWhite100,
     },
     bgDarkMode: {
       100: common.colours.darkGrey100,


app/src/NativeBase/Components/Project/ProjectFullDetails.tsx
```diff
@@ -39,7 +39,7 @@ const ProjectFullDetails = ({
   return (
     <>
       <ScrollView
-        _dark={{ backgroundColor: 'bgDarkMode.100' }}
+        _dark={{ backgroundColor: 'bgDarkMode.200' }}
```
joefhall, line 42: Could we actually remove these two lines here (42 and 43), and remove the `ScrollView` default background colour settings in `StaTheme.tsx`?  

I can't see any difference between the background colour being used here and the overall app background colours set in `navigationTheme` in `Application.tsx` (and so the `ScrollView` can just be transparent) -- except that there the light background is set to white, and I think that's an accessibility no-no, I'm going to check with the team on Slack.

kelsiesmurphy: Happy to change this if required.

joefhall: Thanks @kelsiesmurphy just had confirmation from Joanna on the Jira ticket.  

So I think we should change `bg.100` in `StaTheme` to `offWhite100` and remove `bg.secondary` (as it would be the same).  And remove the `ScrollView` default background colours there.  And then you can remove these two lines in this file too.

joefhall: (And hopefully my Friday brain hasn't missed anything / suggested anything daft there!)

DougieWougie (approved):
Nice one @kelsiesmurphy! I agree with @joefhall - run it past Mia before you merge but the system/dark/light switch in settings has the same colouring in darkmode as all/saved/my so might be better as a separate ticket?

kelsiesmurphy:
Thanks @joefhall and @DougieWougie. Made the updates on Joe's comment, please check commits. I also spoke to Mia regarding the other part, and this should be updated also. I'm going to set it as another ticket however, as @DougieWougie says, it is another UI change that would affect this tab bar, the settings tab bar, and the search tab bar.

If you're happy with that Joe, I'm happy to merge before the weekend, and i'm happy to pick up the tab bar on Monday if nobody else asks to do it.

app/src/NativeBase/Components/Project/ProjectFullDetails.tsx
kelsiesmurphy: This is done, over two [commits](https://github.com/Scottish-Tech-Army/Volunteer-app/pull/155/commits/09edb9ebb7f60361bea0082ffc907bd9cd99c45a). I've added a longer form comment below, but I did test this, and I can't see any areas where that secondary bg value was being referenced in the app, but since it is now removed there's a possibility somewhere is sneakily referencing it.

kelsiesmurphy:
Awesome! Jira ticket for the next part has been created under SVA-307.

PR #186, opened 2023-07-13
Feature/SVA-354-my-experience-profile-screen

DJMF2015:
- [Description](#description)
- [Checklist](#checklist)

# Description

## Type of change
 
# Checklist

- [X] I have checked the app in dark mode, if making front-end design changes


M4RC1N76:
Hi @DJMF2015 , just a quick question, I have had a: --->  WARN  ❗  You are running in development mode but you are using the production API server -- this is usually a bad idea, you should be using your local API development server instead. also When I run the Vol App in Expo Go and went to the "My Profile" there was a blank screen. could you tell is it my android having some issues or is it something else?

M4RC1N76:
@DJMF2015 , nice one!! I did manage to go to the experience and tick for example  "web developer" or other boxes. I am aware that your ticket is inter-linked with Pamela's ticket and that there will be more functions added once her ticket is approved. Great work David!!

M4RC1N76 (approved):
Great work David!!

DJMF2015:
@M4RC1N76 for the feedback. I just added a few changes for dark mode... I think one issue (s) may be similar to what Joe commented on for Pam's ticket about the negative margining but will see what happens..! Cheers

M4RC1N76:
@DJMF2015 I did tested your PR in the dark mode too. It seemed okay. Today I will check the progress on Pamela's ticket since yesterday. I'll speak to you later David.
Regards Marcin Borowski

Sent from Yahoo Mail on Android

app/src/NativeBase/Containers/MyExperienceContainer.tsx
```diff
@@ -0,0 +1,108 @@
...
+   */
+  const filteredSkills = roleGroups.filter((roleGroup: RoleGroup) => {
+    return roleGroup.groupName.toLowerCase().includes(searchTxt.toLowerCase())
+  })
+
+  return (
+    <>
+      <HStack marginTop={-10} marginBottom={12} alignItems={'flex-end'}>
+        <Heading size="md" fontWeight="bold" marginRight={-5}>
+          My Experience
```
PammyLo, line 35: If I'm looking at the correct design I think this needs to be the `BebasNeue-Regular` font. I recall we added it together when we did the code fly through. Is there a reason you removed it from `StaTheme`?

app/src/NativeBase/Containers/MyExperienceContainer.tsx
```diff
@@ -0,0 +1,108 @@
...
+                value={roleGroup.groupName}
+                key={`${roleGroup.groupName} - ${index}`}
+              >
+                {roleGroup.groupName}
+              </Checkbox>
+            ))}
+          </Checkbox.Group>
+        </VStack>
+        <VStack marginX={6}>
+          <Button
```
PammyLo, line 77: I'm getting an error `We can not support a function callback` when I select a checkbox. 

Also the checkbox turns blue when selected, think this should be purple on this page? Or just a tick if possible as per the design, but I suspect the NativeBase checkboxes don't behave like that out of the box.  

Also not sure what the desired behaviour is in terms of static content and content that scrolls (ie should the search box or the next buttons perhaps stay on screen when scrolling?)

app/src/NativeBase/Containers/MyExperienceContainer.tsx
```diff
@@ -0,0 +1,108 @@
...
+    <>
+      <HStack marginTop={-10} marginBottom={12} alignItems={'flex-end'}>
+        <Heading size="md" fontWeight="bold" marginRight={-5}>
+          My Experience
+        </Heading>
+        <Flex flex={1} alignItems={'flex-end'}>
+          <StaThemeLogo />
+        </Flex>
+      </HStack>
+      <StaRibbon />
```
PammyLo, line 41: Great component David. It looks like this may not be needed here for this screen though, as it seems to be replaced with a progress bar in the design.

app/src/NativeBase/Containers/MyExperienceContainer.tsx
```diff
@@ -0,0 +1,108 @@
...
+  })
+
+  return (
+    <>
+      <HStack marginTop={-10} marginBottom={12} alignItems={'flex-end'}>
+        <Heading size="md" fontWeight="bold" marginRight={-5}>
+          My Experience
+        </Heading>
+        <Flex flex={1} alignItems={'flex-end'}>
+          <StaThemeLogo />
```
PammyLo, line 38: The logo seems longer than the design, I wonder if you could push it up a bit?  One way to do that could be to limit the height and hide the overflow.   

Also, on my screen (iPhone) there appears to be a little too much white space under the title.

PammyLo (changes requested):
Hi David, great work! 
I've added a few suggestions and questions in the code. Once the ProfileContainer has been updated and merged will have a more in-depth review :)

app/src/NativeBase/Containers/MyExperienceContainer.tsx
DJMF2015: Hi Pamela, I will address other comments tomorrow but on the issue of the warning about a 'function callback ' I tried a number of fixes from googling about this issue. It seems to be an error that comes up in Nativebase from researching but haven't had any luck finding a fix yet. Will leave this until other issues are sorted. I'm terms of the buttons and search it's difficult to know if they are meant to remain fixed when scrolling as nothing on the designs to indicate but I presumed not.

Thanks for reviewing and will come back about the other comments...

app/src/NativeBase/Containers/MyExperienceContainer.tsx
DJMF2015: According to Joanna this was to he the STA Ribbon instead which is why I have it included in place of the other object.

app/src/NativeBase/Containers/MyExperienceContainer.tsx
DJMF2015: @PammyLo I have reduce the whitespace between the Heading and searchbar. As for the logo it looks fine on my mobiles but when i tried what you suggested the image becomes skewed.

app/src/NativeBase/Containers/MyExperienceContainer.tsx
DJMF2015: Just to add to the comment above, Joanna was clear she wanted the STA ribbons included but leaves less room for the progress bar. So need to clarify this point. If using the progress bar, Nativebase has one but looks considerably different to Mia's designs 'out-of-the-box'

app/src/NativeBase/Containers/MyExperienceContainer.tsx
DJMF2015: @PammyLo I have now made the searchbar fixed so that it remains in plave when the user scolls down. However i have not done this for the two buttons at the bottom of the screen as felt need more clarity from Mia on this. It makes sense  for the searchbar which is why i have made the suggestion you described above for this. As for the 'function callback' error i am no nearer resolving the issue. After research, it does appear a bug but with as yet no obvious fix for it. This was as recent as 3 months ago see https://github.com/GeekyAnts/NativeBase/issues/5098

app/src/NativeBase/Containers/MyExperienceContainer.tsx
DJMF2015: @PammyLo i lost the changes we made when i switched from the main branch. I have tried incorporating the above font but no luck yet getting it working. Obviously i am doing somethign wrong . WIll have another go later on this afternoon.

app/src/NativeBase/Components/StaRibbon.tsx
```diff
@@ -0,0 +1,20 @@
+/**
+ * @file for StaRibbon component
+ */
+import { HStack, Box } from 'native-base'
+import React from 'react'
+
+const StaRibbon = () => {
```
joefhall, line 7: Optional little bit of shorthand to make this file simpler, because this component just returns JSX without any other logic:
you can make this line `const StaRibbon = () => (`, remove the following line, and remove the closing curly bracket on the penultimate line of the file.

app/src/NativeBase/Containers/MyExperienceContainer.tsx
```diff
@@ -0,0 +1,117 @@
...
+ */
+import {
+  Heading,
+  VStack,
+  HStack,
+  Checkbox,
+  ScrollView,
+  Flex,
+  Button,
+  Box,
```
joefhall, line 12: It looks like `Box` isn't being used so this import can be removed here

app/src/NativeBase/Containers/MyExperienceContainer.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { RoleGroup, roleGroups } from '@/Services/modules/projects/roleGroups'
+
+const MyExperienceContainer = () => {
+  const [skillsValue, setSkillsValue] = React.useState<string[]>([])
+  const [searchTxt, setSearchTxt] = React.useState<string>('') // search text
+
+  /**
+   * Filter the role groups based on the search text entered by the user
+   */
+  const filteredSkills = roleGroups.filter((roleGroup: RoleGroup) => {
```
joefhall, line 28: Another optional bit of shorthand you could use here -- this function could be simplified to this (because it's a one-line function we don't have to use curly brackets and `return`):

```
const filteredSkills = roleGroups.filter((roleGroup: RoleGroup) =>
  roleGroup.groupName.toLowerCase().includes(searchTxt.toLowerCase()),
)
```

app/src/NativeBase/Containers/MyExperienceContainer.tsx
```diff
@@ -0,0 +1,117 @@
...
+        </Heading>
+        <Flex flex={1} alignItems={'flex-end'} mt={-10}>
+          <StaThemeLogo width={'100%'} />
+        </Flex>
+      </HStack>
+      <StaRibbon />
+      <View margin={'4'}>
+        <FreeSearchBar
+          marginTop="2"
+          handleSubmit={() => {
```
joefhall, line 53: Could simplify these three lines to be just `handleSubmit={() => null}`

app/src/NativeBase/Containers/MyExperienceContainer.tsx
```diff
@@ -0,0 +1,117 @@
...
+            null
+          }}
+          handleChangeText={(text: string) => {
```
joefhall, line 56: And these three lines could be simplified to just `handleChangeText={(text: string) => setSearchTxt(text)}` (again, just a bit of a handy shorthand for one-line functions)

app/src/NativeBase/Containers/MyExperienceContainer.tsx
```diff
@@ -0,0 +1,117 @@
...
+            setSearchTxt(text)
+          }}
+        />
+      </View>
+      <ScrollView>
+        <VStack mt={-3} margin="2">
+          <Checkbox.Group
+            ml="2"
+            onChange={(value: string[]) => {
```
joefhall, line 65: Similar point to above -- could simplify this into a one-liner without curly brackets around the function

app/src/NativeBase/Containers/MyExperienceContainer.tsx
joefhall: Wow that is annoying about that NativeBase warning -- ages without a fix for a very basic component (slightly terrifying!)

Could either use [a different checkbox](https://github.com/react-native-checkbox/react-native-checkbox) (but it might need to be styled) -- or we just ignore it?  I'm not sure whether it'll show up in Bugsnag (our bug logging service) or not as it's a warning.

app/src/NativeBase/Containers/MyExperienceContainer.tsx
```diff
@@ -0,0 +1,117 @@
...
+        >
+          My Experience
+        </Heading>
+        <Flex flex={1} alignItems={'flex-end'} mt={-10}>
+          <StaThemeLogo width={'100%'} />
+        </Flex>
+      </HStack>
+      <StaRibbon />
+      <View margin={'4'}>
+        <FreeSearchBar
```
joefhall, line 51: Would it be possible to make it so the user can come out of the free search input by tapping anywhere else on the screen?  The only way you can do that at the moment is to tap on one of the checkboxes.

I started searching and then wanted to see the full list of results, but some of the results were hidden underneath my keyboard.  If we could make it possible to 'blur' / lose focus from the text input by tapping elsewhere on the screen background that would make it possible to see all the checkboxes in the results list.

joefhall (changes requested):
Nice work on this David -- looking great.

I had one question about being able to come out of the free search box to see the checkboxes below (without selecting one), otherwise just a few small stylistic suggestions.

app/src/NativeBase/Containers/MyExperienceContainer.tsx
DJMF2015: Hi Joe, thanks for all the comments and very helpful. Will see to those changes  tomorrow. Just on the point about the search box I think I understand what you're asking but wanted to clarify. Are you meaning that when a user has typed something into the searchbox e.g. ''web developer" and then clicks _anywhere_ else on the screen (loses focus), all the checkboxes should '_reappear_' again with the searchbox being 'cleared'?

app/src/NativeBase/Components/StaRibbon.tsx
DJMF2015: I have now removed the unnecessary curly braces from the sta ribbon.

app/src/NativeBase/Containers/MyExperienceContainer.tsx
DJMF2015: I have reduced this to the one line as suggested.

app/src/NativeBase/Containers/MyExperienceContainer.tsx
DJMF2015: As above, have reduced the 3 lines to the 1 line by removing the curly brackets where not required.

app/src/NativeBase/Containers/MyExperienceContainer.tsx
joefhall: Hi @DJMF2015 are you on an Android phone? I’m on an iPhone, I wonder if that could be the difference? When I tap outside the search box it doesn’t lose focus (the keyboard remains) — unless I tap on a checkbox. Here’s how it looks for me in this video

https://github.com/Scottish-Tech-Army/Volunteer-app/assets/25086222/c6f95da3-30df-4012-8c08-564b9ae96491

joefhall: (The part where nothing seems to be happening is me tapping on different parts of the screen)

DJMF2015: @joefhall thanks for providing that. I think it may well be an iphone issue as i am on android and it doesn't do it for me (when i click outside the text box the keyboard automatically disappears). I read there is something called `TouchableWithoutFeedback` which allows the keyboard to be dismissed when they user loses focus? Perhaps i can add this and see if it makes a difference. `TouchableWithoutFeedback` is a react-native specific module.

DJMF2015: @joefhall just added an attempt at a fix for the problem you are experiencing on ios. Let me know if it makes any difference as i only have android myself.

app/src/NativeBase/Containers/MyExperienceContainer.tsx
PammyLo: Ok, cool, thanks David, we'll see where the design goes, in any case great work :)

app/src/NativeBase/Containers/MyExperienceContainer.tsx
PammyLo: If you pull the new changes into main and merge into your branch then the theme will be up to date. You could then use fontFamily=“title” here and you won’t need fontWeight.

It does work this way but the idea is that the fonts are set in the theme, so that should they change in the future it only needs to be updated in the theme.

app/src/NativeBase/Containers/MyExperienceContainer.tsx
PammyLo: Makes sense to ask Mia what the intended behaviour is of the search and buttons. I don't know either, was more a question than a suggestion. 

I know the checkbox from Native Base is hard to change from solid to just a tick, possibly not a problem I'd say but also up to Mia or Joanna. Making it purple would be good though. 

That error really is annoying. I'll leave it up to you and Joe to decide how to deal with that ;)

app/src/NativeBase/Containers/MyExperienceContainer.tsx
PammyLo: Hi David, it's looking really great. Just on my phone I can't see the last two colours, yellow and red, and recall Joanna saying all 7 should be visible. It's a CSS challenge for sure..
The whitespace above the search bar is now fine but there still seems to be a lot between title and STA ribbon.

DJMF2015: Hi Pamela that's unfortunate. Iam not having a lot with this ticket. I suspect it's down to the screen size you are using and as is why you can't view all the colours of the logo. I am not honestly sure how I can resolve this issue as each time I change  something to the logo it causes another issue in sizing.

app/src/NativeBase/Containers/MyExperienceContainer.tsx
DJMF2015: Thanks for that Pamela! I will pull the branch shortly and merge your changes and then hopefully have the font from the theme like you suggest.

app/src/NativeBase/Containers/MyExperienceContainer.tsx
joefhall: Thanks @DJMF2015 unfortunately not -- it's still the same for me

joefhall: @DJMF2015 I've found a way that works on iPhone -- if you make the top-level element in MyExperienceContainer `<Pressable onPress={() => Keyboard.dismiss()}>` (with `</Pressable>` the very last line of what's returned).  (You'll need to import `Pressable` from NativeBase.)

Fingers crossed that works on Android too.

app/src/NativeBase/Components/Checkbox.tsx
```diff
@@ -0,0 +1,44 @@
+/**
+ * @file CheckBox component for the STA Volunteer App
+ */
+
+import { Pressable } from 'native-base'
+import React from 'react'
+import { Ionicons } from '@expo/vector-icons'
```
joefhall, line 7: Was there a reason not to use a Material Icons icon here?  The one you've used looks fine, but just for consistency with the rest of the project https://github.com/Scottish-Tech-Army/Volunteer-app/blob/develop/APP_DEVELOPMENT.md#icons

app/src/NativeBase/Components/Checkbox.tsx
```diff
@@ -0,0 +1,44 @@
...
+ * @param {value: string[] => void} props.onChange function called when the checkbox is pressed
+ * @param {boolean} props.checked whether the checkbox is checked or not
+ * @returns {React.ReactElement} Component
+ */
+
+const Checkbox: React.FC<CheckboxProps> = ({ onChange, checked }) => {
+  return (
+    <Pressable
+      onPress={() => onChange([])}
+      flexDirection={'row'}
```
joefhall, line 27: Tiny thing, but from looking at examples in NativeBase docs, I think normal style for setting text properties here and below is just to use double quotes e.g. `flexDirection="row"` (see e.g. examples at https://docs.nativebase.io/pressable)

app/src/NativeBase/Components/Checkbox.tsx
```diff
@@ -0,0 +1,44 @@
...
+      borderWidth={2}
+      borderRadius={2}
+      width={30}
+      height={30}
+      justifyContent={'center'}
+      _dark={{
+        borderColor: 'white',
+      }}
+    >
+      {checked && <Ionicons name="checkmark" size={20} color="#D1338A" />}
```
joefhall, line 39: Please always use colours from the theme rather than putting hex values directly into a component -- this makes it much easier to keep consistency and make updates later on.

If it won't let you reference the colour directly here as `color="magenta.100"` you can import the StaTheme file and reference the colour value from there (see e.g. [here](https://github.com/Scottish-Tech-Army/Volunteer-app/blob/8ee48822ba6fa2edbf9b57fcef33c21a39e78b46/app/src/NativeBase/Components/TopOfApp.tsx#L55))

joefhall: Also would it be possible to use the NativeBase `<Icon />` component here, like we usually do for icons, or was there a reason not to?  https://github.com/Scottish-Tech-Army/Volunteer-app/blob/develop/APP_DEVELOPMENT.md#icons

app/src/NativeBase/Components/ProfileButtons.tsx
```diff
@@ -0,0 +1,32 @@
+/**
```
joefhall, line 1: I'm a bit confused as to why we need this component, when it basically just returns a button?  It doesn't do anything else apart from passing on various props -- so is there a reason not to just directly use a `<Button />` instead and remove this component?

app/src/NativeBase/Components/Checkbox.tsx
```diff
@@ -0,0 +1,44 @@
+/**
+ * @file CheckBox component for the STA Volunteer App
+ */
+
+import { Pressable } from 'native-base'
+import React from 'react'
+import { Ionicons } from '@expo/vector-icons'
+
+type CheckboxProps = {
+  onChange: (value: string[]) => void
```
joefhall, line 10: Why does a checkbox return a string array?  I would have expected each checkbox to just return a string?

app/src/NativeBase/Components/Checkbox.tsx
```diff
@@ -0,0 +1,44 @@
...
+ * @param { CheckboxProps} props The component props
+ * @param {value: string[] => void} props.onChange function called when the checkbox is pressed
+ * @param {boolean} props.checked whether the checkbox is checked or not
+ * @returns {React.ReactElement} Component
+ */
+
+const Checkbox: React.FC<CheckboxProps> = ({ onChange, checked }) => {
+  return (
+    <Pressable
+      onPress={() => onChange([])}
```
joefhall, line 26: Why does `onChange` take a `value` parameter if it doesn't do anything with it?  If it's not actually being used, could we just remove this and then this line could simply be `onPress={onChange}`?

(I was also confused as to why a single checkbox would return an array of strings, not just a string)

app/src/NativeBase/Containers/MyExperienceContainer.tsx
```diff
@@ -0,0 +1,141 @@
...
+      <HStack mt={-10} mb={5} alignItems={'flex-end'}>
+        <Heading size="md" fontFamily="title" adjustsFontSizeToFit>
+          My Experience
+        </Heading>
+        <Flex flex={1} mt={-5} alignItems={'flex-end'}>
+          <StaThemeLogo />
+        </Flex>
+      </HStack>
+      <ProgressBar colours={progressBarColourObject} />
+      <View margin={'4'}>
```
joefhall, line 53: Tiny thing, but there are some minor inconsistencies in how values are being set -- it's not entirely clear from the NativeBase docs, but it looks like for number settings either `margin="4"` or `margin={4}` are both valid

app/src/Utils/ProgressBarColours.ts
```diff
@@ -0,0 +1,54 @@
...
+ * @param currentBox
+ * @param bgColor
+ * @returns  {ProgressBarColors} - returns the colours for the progress bar
+ */
+
+export const getProgressBarColors = (
+  currentBox: number,
+  bgColor: string,
+): ProgressBarColors => {
+  const boxes = ['#F1F1F1', '#F1F1F1', '#F1F1F1'] // default colour for all boxes while the user is on the same screen
```
joefhall, line 24: Please can this colour be defined in StaTheme and then called from there, rather than using hex values anywhere in this file

app/src/Utils/ProgressBarColours.ts
```diff
@@ -0,0 +1,54 @@
...
+  for (let i = 0; i < currentBox; i++) {
+    boxes[i] = bgColor
+  }
+
+  return {
+    box1: boxes[0],
+    box2: boxes[1],
+    box3: boxes[2],
+    bgColor: boxes[3] === bgColor ? bgColor : '#F1F1F1',
```
joefhall, line 33: Tiny thing, but as good practice could use the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) here to simplify this to `bgColor: boxes[3] === bgColor ?? '#F1F1F1',`

(but obviously update the colour reference to call it from StaTheme)

app/src/NativeBase/Containers/MyExperienceContainer.tsx
```diff
@@ -0,0 +1,141 @@
...
+                  }
+                }}
+              >
+                {roleGroup.groupName}
+              </Text>
+            </View>
+          ))}
+        </VStack>
+        <VStack marginX={5}>
+          <ProfileButtons
```
joefhall, line 106: It's my bad for not including it in the list of custom components [here](https://github.com/Scottish-Tech-Army/Volunteer-app/blob/develop/APP_DEVELOPMENT.md#custom-components) (sorry about that), but there's already a `SecondaryButton` component which I think might be styled to do what you need here with the outline?  (so would save you having to set all these settings manually on the button -- and more reusable)

joefhall: Also, from looking at Mia's designs on Figma, is the idea that the button has the outline until at least one checkbox is ticked, and then it becomes solid purple?  (i.e. the outlined version is a disabled state look)  Is getting the button to change that way within the scope of this ticket?

joefhall (changes requested):
Nice work @DJMF2015 I've added some questions and suggestions below, hope these make sense and are useful.  

On one or two things I'm not sure about the scope of the ticket you're working on, so will be steered by you on that.

app/src/NativeBase/Containers/MyExperienceContainer.tsx
DJMF2015: Thanks @joefhall. Wasn't aware of the SecondaryButton. Will look into this change.

app/src/Utils/ProgressBarColours.ts
DJMF2015: @joefhall good point - will do!

app/src/NativeBase/Components/Checkbox.tsx
DJMF2015: Thanks . I wasn't aware of this Icon component. Will look into.

app/src/NativeBase/Components/Checkbox.tsx
DJMF2015: Used the above as came with expo-checbox. I didn't appreciate we used MaterialIcons but i can look into swapping this out.

app/src/NativeBase/Components/Checkbox.tsx
DJMF2015: I have updated the Pressable properties to use double-quotes instead.

app/src/NativeBase/Components/Checkbox.tsx
DJMF2015: I have added in the Material Icon with checkbox 'tick'. I have imported the StaTheme file to utilise the 'magenta.100' colour.

app/src/Utils/ProgressBarColours.ts
DJMF2015: I have updated these to use the StaTheme file colours.

app/src/NativeBase/Components/Checkbox.tsx
```diff
@@ -0,0 +1,51 @@
...
+type CheckboxProps = {
+  onChange: () => void
+  checked: boolean
+}
+
+/**
+ * Component that displays a checkbox that can be checked or unchecked.
+ *
+ * @param { CheckboxProps} props The component props
+ * @param {value: string[] => void} props.onChange function called when the checkbox is pressed
```
joefhall, line 18: Ideally this would be corrected to `() => void` (rather than `string[] => void`)

app/src/NativeBase/Components/Forms/DatePicker.tsx
```diff
@@ -109,7 +109,7 @@ const DatePicker: FC<DatePickerProps> = ({
           <Text
             paddingTop="8"
             fontSize="sm"
-            fontFamily="primarySemiBold"
+            fontWeight="600"
```
joefhall, line 112: With all these changes from `fontFamily` to a `fontWeight` setting, Pam only just changed these -- assume you've checked with her or have a good reason to change them again?

joefhall (approved):
Cheers David.

Have approved, and left a couple of comments on things you might want to double-check -- but I'll leave it to you at this point.

app/src/NativeBase/Components/Checkbox.tsx
DJMF2015: Apologies, thought I had addressed this. Will get it changed. Cheers!

app/src/NativeBase/Components/Forms/DatePicker.tsx
joefhall: Your changes here are reversing some changes Pam made to fonts -- e.g. here you can see how it's set on the `develop` branch: https://github.com/Scottish-Tech-Army/Volunteer-app/blob/develop/app/src/NativeBase/Components/Forms/DatePicker.tsx#L112

PammyLo (approved):
Great work David @DJMF2015!

PR #187, opened 2023-07-13
feature/SVA-386 add placeholder content to Profile screen container

PammyLo:
- [Description](#description)
- [Checklist](#checklist)

# Description

- This adds content to the Profile screen container, with placeholder content for user details
- menu items ready to be connected to new screens 

Fixes # (SVA-386)[https://sta2020.atlassian.net/browse/SVA-386]

## Type of change

- [] Bug fix (non-breaking change which fixes an issue)

# Testing locally

- run locally and check it looks as per design

# Checklist

- [X] I have checked the app in dark mode, if making front-end design changes
- [X] New and existing unit tests pass locally with my changes

### DIFF
--- a/app/src/NativeBase/Components/ChoicesList.tsx
+++ b/app/src/NativeBase/Components/ChoicesList.tsx
@@ -8,15 +8,45 @@ import MaterialIcons from 'react-native-vector-icons/MaterialIcons'
 
 export interface ChoicesListChoice {
   text: string
-  color: string
-  darkModeColor: string
-  fontSize?: string
-  fontWeight?: string
   onPress: () => void
+}
+
+export enum ChoicesListColour {
+  primary = 'primary',
+  purple = 'purple',
+}
+
+const colourStyles = {
+  [ChoicesListColour.primary]: {
+    color: 'primary.100',
+    darkModeColor: 'primary.40',
+  },
+  [ChoicesListColour.purple]: {
+    color: 'purple.100',
+    darkModeColor: 'purple.40',
+  },
+}
+
+export enum ChoicesListFontStyle {
+  mediumLight = 'mediumLight',
+  smallSemiBold = 'smallSemiBold',
+}
+
+const fontStyles = {
+  [ChoicesListFontStyle.mediumLight]: {
+    fontSize: 'md',
+    fontWeight: '300',
+  },
+  [ChoicesListFontStyle.smallSemiBold]: {
+    fontSize: 'sm',
+    fontWeight: '600',
+  },
 }
 
 interface ChoicesListProps {
   choices: ChoicesListChoice[]
+  colour: ChoicesListColour
+  style: ChoicesListFontStyle
 }
 
 /**
@@ -26,7 +56,7 @@ interface ChoicesListProps {
  * @param {ChoicesListChoice[]} props.choices The choices to show in the list
  * @returns {React.ReactElement} Component
  */
-const ChoicesList: FC<ChoicesListProps> = ({ choices }) => (
+const ChoicesList: FC<ChoicesListProps> = ({ choices, colour, style }) => (
   <VStack>
     {choices.map(choice => (
       <Pressable key={choice.text} onPress={choice.onPress}>
@@ -36,13 +66,16 @@ const ChoicesList: FC<ChoicesListProps> = ({ choices }) => (
           borderBottomWidth="1"
           justifyContent="space-between"
         >
-          <Text fontWeight={choice.fontWeight} fontSize={choice.fontSize}>
+          <Text
+            fontWeight={fontStyles[style].fontWeight}
+            fontSize={fontStyles[style].fontSize}
+          >
             {choice.text}
           </Text>
           <Icon
             as={MaterialIcons}
-            color={choice.color}
-            _dark={{ color: choice.darkModeColor }}
+            color={colourStyles[colour].color}
+            _dark={{ color: colourStyles[colour].darkModeColor }}
             name="arrow-forward-ios"
             size="md"
           />
--- a/app/src/NativeBase/Components/ChoicesList.tsx
+++ b/app/src/NativeBase/Components/ChoicesList.tsx
@@ -8,6 +8,10 @@ import MaterialIcons from 'react-native-vector-icons/MaterialIcons'
 
 export interface ChoicesListChoice {
   text: string
+  color: string
+  darkModeColor: string
+  fontSize?: string
+  fontWeight?: string
   onPress: () => void
 }
 
@@ -32,10 +36,13 @@ const ChoicesList: FC<ChoicesListProps> = ({ choices }) => (
           borderBottomWidth="1"
           justifyContent="space-between"
         >
-          <Text fontWeight="300">{choice.text}</Text>
+          <Text fontWeight={choice.fontWeight} fontSize={choice.fontSize}>
+            {choice.text}
+          </Text>
           <Icon
             as={MaterialIcons}
-            color="primary.100"
+            color={choice.color}
+            _dark={{ color: choice.darkModeColor }}
             name="arrow-forward-ios"
             size="md"
           />
--- a/app/src/NativeBase/Components/ChoicesList.tsx
+++ b/app/src/NativeBase/Components/ChoicesList.tsx
@@ -35,11 +35,11 @@ export enum ChoicesListFontStyle {
 const fontStyles = {
   [ChoicesListFontStyle.mediumLight]: {
     fontSize: 'md',
-    fontWeight: '300',
+    fontFamily: 'primaryLight',
   },
   [ChoicesListFontStyle.smallSemiBold]: {
     fontSize: 'sm',
-    fontWeight: '600',
+    fontFamily: 'primarySemiBold',
   },
 }
 
@@ -67,7 +67,7 @@ const ChoicesList: FC<ChoicesListProps> = ({ choices, colour, style }) => (
           justifyContent="space-between"
         >
           <Text
-            fontWeight={fontStyles[style].fontWeight}
+            fontFamily={fontStyles[style].fontFamily}
             fontSize={fontStyles[style].fontSize}
           >
             {choice.text}
--- a/app/src/NativeBase/Containers/ProfileContainer.tsx
+++ b/app/src/NativeBase/Containers/ProfileContainer.tsx
@@ -2,8 +2,39 @@
  * @file Profile screen showing user's personal profile.
  */
 
-import { Heading, VStack, Center } from 'native-base'
+import { AntDesign, Feather } from '@expo/vector-icons'
+import {
+  Heading,
+  VStack,
+  Center,
+  Icon,
+  HStack,
+  Box,
+  Text,
+  ScrollView,
+} from 'native-base'
 import React from 'react'
+import ChoicesList, { ChoicesListChoice } from '../Components/ChoicesList'
+import underDevelopmentAlert from '../../Utils/UnderDevelopmentAlert'
+
+const profileMenuItems: string[] = [
+  'My Projects',
+  'My Experience',
+  'Volunteer Vibes',
+  'Notifications & Settings',
+]
+
+const profileMenuChoices = profileMenuItems.map(
+  (item: string) =>
+    ({
+      text: item,
+      color: 'purple.100',
+      darkModeColor: 'purple.40',
+      fontSize: 'sm',
+      fontWeight: '600',
+      onPress: underDevelopmentAlert,
+    } as ChoicesListChoice),
+)
 
 /**
  * A screen container that displays the user's profile.
@@ -13,20 +44,72 @@ import React from 'react'
 const ProfileContainer = () => {
   return (
     <VStack
-      paddingTop="4"
-      paddingBottom="2"
-      alignItems="center"
-      space="4"
-      backgroundColor="secondaryGrey.60"
+      padding={2}
+      backgroundColor="bg.100"
       _dark={{ backgroundColor: 'bgDarkMode.100' }}
     >
-      <Center
-        backgroundColor="bg.100"
-        _dark={{ backgroundColor: 'bgDarkMode.100' }}
-        width="100%"
-      >
-        <Heading size="sm">My Profile</Heading>
+      <Center>
+        <Heading fontFamily="title" size="lg">
+          My Profile
+        </Heading>
+        <Box
+          marginTop={6}
+          height="93px"
+          width="90px"
+          borderColor="darkGrey.100"
+          borderWidth="4px"
+          borderRadius="full"
+        />
+        <HStack>
+          <Heading fontSize="md" fontWeight="700">
+            Full Name
+          </Heading>
+          <Text fontSize="xs" paddingLeft="-2">
+            (She)
+          </Text>
+        </HStack>
+        <Text fontSize="xs" marginTop="-4">
+          P-R-O-N-O-U-N-C-I-A-T-I-O-N
+        </Text>
+        <Text fontSize="xs" paddingTop={-2}>
+          Role
+        </Text>
+        <Text fontSize="xs" paddingTop={-2}>
+          Status
+        </Text>
+        <Heading fontSize="md" fontWeight="700" paddingTop={-2}>
+          My volunteering
+        </Heading>
+        <Box
+          bg="secondaryGrey.60"
+          _dark={{ bg: 'bgDarkMode.200' }}
+          paddingLeft={4}
+          shadow="9"
+          marginTop={2}
+          padding={2}
+          width="92%"
+        >
+          <HStack>
+            <Icon
+              as={Feather}
+              color="darkGrey.100"
+              name="award"
+              size={6}
+              marginTop={2}
+            />
+            <Text fontSize="sm">Completed projects</Text>
+          </HStack>
+          <HStack>
+            <Icon as={AntDesign} color="darkGrey.100" name="rocket1" size={6} />
+            <Text fontSize="sm" paddingTop={-2}>
+              Ongoing projects
+            </Text>
+          </HStack>
+        </Box>
       </Center>
+      <ScrollView marginTop={2}>
+        <ChoicesList choices={profileMenuChoices} />
+      </ScrollView>
     </VStack>
   )
 }
--- a/app/src/NativeBase/Containers/ProfileContainer.tsx
+++ b/app/src/NativeBase/Containers/ProfileContainer.tsx
@@ -13,6 +13,7 @@ import {
   Heading,
   HStack,
   Icon,
+  Pressable,
   ScrollView,
   Text,
   VStack,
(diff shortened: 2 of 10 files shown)


M4RC1N76:
Hi @PammyLo , I have run the Vol App and went to the Profile section. The page Looks nearly like the Design (on my android phone). Can you see whole list of titles to choose from i.e. Roles & Interests, STA activities, Notifications, Share app and Volunteering journey? In my case I can only see  "Roles & Interests" and "STA activities". I could not scroll down to the bottom of the list to see the all titles like on the "My Profile" design, could you let me know if this is my android issue or something else, please?

PammyLo:
> Hi @PammyLo , I have run the Vol App and went to the Profile section. The page Looks nearly like the Design (on my android phone). Can you see whole list of titles to choose from i.e. Roles & Interests, STA activities, Notifications, Share app and Volunteering journey? In my case I can only see "Roles & Interests" and "STA activities". I could not scroll down to the bottom of the list to see the all titles like on the "My Profile" design, could you let me know if this is my android issue or something else, please?

That is interesting @M4RC1N76, that you see "Roles & Interests" and "STA activities"... The screen should look like the image below, more or less that is as this is an iPhone screenshot so the very top and that bar at the bottom will look slightly different on Android, but the menu items should be the same and should be scrollable if your screen is smaller. I wonder if you are in another branch? 

image

app/src/NativeBase/Components/ChoicesList.tsx
DJMF2015, line 1: This is all great!

app/src/Navigators/Main.tsx
DJMF2015, line 1: This is fine. I did a similiar thing with the ProfileContainer for MyExperience so as to be able to have a way to navigate to it. Once this is merged  the MyExperience contaienr can use this.

app/src/NativeBase/Theme/StaTheme.tsx
DJMF2015, line 1: Great use of fonts

app/src/NativeBase/Containers/ProfileContainer.tsx
DJMF2015, line 1: This all looks fine to me and good use of the vector icons as well as being clean and consistent.

app/src/Navigators/Main.tsx
DJMF2015: @PammyLo I have review your code and importantly too, tried the app on my phone. It looks absolutely fine in regards to matching the figma design. I assume the 'padlock' icon which is missing is something that can be added later if necessary? I don't have any issues running it on my phone and everything is well laid out and no bugs.

DJMF2015 (commented):
@PammyLo good work. So my comments but looks fine to me and works properly on my mobile phone without bugs. Happy to approve.

app/src/NativeBase/Containers/ProfileContainer.tsx
M4RC1N76: Thanks @PammyLo . I was in the right branch. I'll check couple of other things and approve when tested. I am aware that @DJMF2015 is working on the ticket inter-linked with yours so it would be good to have those two approved :)

app/src/NativeBase/Containers/ProfileContainer.tsx
```diff
@@ -13,20 +44,72 @@ import React from 'react'
...
+        />
+        <HStack>
+          <Heading fontSize="md" fontWeight="700">
+            Full Name
+          </Heading>
+          <Text fontSize="xs" paddingLeft="-2">
+            (She)
+          </Text>
+        </HStack>
+        <Text fontSize="xs" marginTop="-4">
```
joefhall, line 71: It looks like there are several negative `paddingTop` and `marginTop` values here -- is there a better alternative to having to set this repeatedly for each element?  E.g. could the `Center` be a `VStack` instead and then set the `spacing` property on that?

app/src/NativeBase/Containers/ProfileContainer.tsx
```diff
@@ -13,20 +44,72 @@ import React from 'react'
...
-        width="100%"
-      >
-        <Heading size="sm">My Profile</Heading>
+      <Center>
+        <Heading fontFamily="title" size="lg">
+          My Profile
+        </Heading>
+        <Box
+          marginTop={6}
+          height="93px"
```
joefhall, line 57: Is there a reason the width and height are different values, if it's a circle?

app/src/NativeBase/Components/ChoicesList.tsx
```diff
@@ -8,6 +8,10 @@ import MaterialIcons from 'react-native-vector-icons/MaterialIcons'
 
 export interface ChoicesListChoice {
   text: string
+  color: string
```
joefhall, line 11: I think these options would be better set for the `ChoicesList` overall rather than for each `ChoicesListChoice` -- because we'd expect all items in one list to share the same colours?  (at least that's what the current designs look like)

joefhall: That would also avoid the need for the `.map` on line 27 of ProfileContainer.tsx, where you're having to repeat the same styling options again and again

joefhall: Actually, just thinking about this some more, this would be more reusable if we didn't have to pass through a whole set of colour options at all, but could just pass through one or two options (e.g. a `colour` e.g. 'purple' and/or a `variant`/`type` property e.g. 'bold') -- and then the colours and font weight etc is set in `ChoicesList` based on that?

That would make it much easier in the future if we want to change e.g. the purple that's used in every 'purple' version of a list.  Or the font sizing, font weight, etc that's used in a 'bold' version of this.  We'd only have to update `ChoicesList`.  Otherwise (the way this is working at the moment) we'd have to find every different place where the ChoicesList is used for a purple/bold list and update all these individual settings possibly many times over.

app/src/NativeBase/Containers/ProfileContainer.tsx
```diff
@@ -13,20 +44,72 @@ import React from 'react'
...
+        <Text fontSize="xs" paddingTop={-2}>
+          Role
+        </Text>
+        <Text fontSize="xs" paddingTop={-2}>
+          Status
+        </Text>
+        <Heading fontSize="md" fontWeight="700" paddingTop={-2}>
+          My volunteering
+        </Heading>
+        <Box
```
joefhall, line 83: Would this work as a `Card` and then it would do things like shadow automatically for you?

app/src/NativeBase/Containers/ProfileContainer.tsx
```diff
@@ -13,20 +44,72 @@ import React from 'react'
...
+          bg="secondaryGrey.60"
+          _dark={{ bg: 'bgDarkMode.200' }}
+          paddingLeft={4}
+          shadow="9"
+          marginTop={2}
+          padding={2}
+          width="92%"
+        >
+          <HStack>
+            <Icon
```
joefhall, line 93: I think these icons need dark mode colours setting -- they're not visible in dark mode

app/src/NativeBase/Containers/ProfileContainer.tsx
```diff
@@ -13,20 +44,72 @@ import React from 'react'
...
-        <Heading size="sm">My Profile</Heading>
+      <Center>
+        <Heading fontFamily="title" size="lg">
+          My Profile
+        </Heading>
+        <Box
+          marginTop={6}
+          height="93px"
+          width="90px"
+          borderColor="darkGrey.100"
```
joefhall, line 59: The border isn't showing up in dark mode

app/src/NativeBase/Theme/StaTheme.tsx
```diff
@@ -28,7 +28,8 @@ const common = {
   },
   fonts: {
     family: {
-      primary: 'Poppins-Regular',
+      primary: 'Poppins',
```
joefhall, line 31: I'm seeing an error message in my terminal: `fontFamily "Poppins" is not a system font and has not been loaded through Font.loadAsync.`

app/src/NativeBase/Containers/ProfileContainer.tsx
```diff
@@ -13,20 +44,72 @@ import React from 'react'
...
+      backgroundColor="bg.100"
       _dark={{ backgroundColor: 'bgDarkMode.100' }}
     >
-      <Center
-        backgroundColor="bg.100"
-        _dark={{ backgroundColor: 'bgDarkMode.100' }}
-        width="100%"
-      >
-        <Heading size="sm">My Profile</Heading>
+      <Center>
```
joefhall, line 51: If the content of this screen ends up being taller than the available space, it doesn't seem to scroll -- content is just cut off e.g. on my phone (iPhone SE 2020) I can't see beyond My Projects and My Experience at the moment.

joefhall (changes requested):
Thanks for this @PammyLo 

I've added some questions/suggestions -- mainly around how to make things reusable for the future and avoid styling being difficult to update, as these are some of the first of many new screens we'll be creating it'll help us if we can get things right at an early stage.  (@DJMF2015 I haven't had a chance to look at your PR yet, but some of these might be relevant there too.)

Also saw some issues with dark mode and shorter mobile screens.

app/src/NativeBase/Containers/ProfileContainer.tsx
DJMF2015: @joefhall yes i have used negative margins on my feature My Experience likewise. I did experiment with using various components but the way i ended up doing this was the only way i could find to get both logo aligned with the Header text and avoid it either disappearing from the screen or wrapping the words one or top of the other. I think if there's a better way to do this on my PR, a few pointers might be helpful given the length of time i spent on it y'day (if that's okay)!

PammyLo: Hi @joefhall, I was counting on you picking this up :) Been wanting to bounce this off you. 

Many of the `marginTop`, `paddingTop` and `paddingLeft` values are actually negative. This is due to the fact that the `Text` component in the `StaTheme` has default padding of 2px. Adding a `space` property of "-2" to the parent component would counteract that, but it would increase the need for additional margin in other places. The result might be more confusing.

The best solution would be to remove the default padding from the `Text` component in `StaTheme` and add padding to the components that contain the text. However that is quite a refactor as it it used throughout the app. Better to do this in a separate ticket? 

Aside from that I have now replaced the `Center` component with `VStack` and `alignItems center` as suggested anyway, as this is probably less confusing and in line with other pages.

app/src/NativeBase/Containers/ProfileContainer.tsx
PammyLo: I've followed the design here, thinking it is intentionally not actually a circle? Happy to change though. Might need to ask Mia?

app/src/NativeBase/Components/ChoicesList.tsx
PammyLo: Ah, excellent idea, much nicer! 💯  Will make those changes, thanks @joefhall! :)

app/src/NativeBase/Containers/ProfileContainer.tsx
```diff
@@ -13,20 +44,72 @@ import React from 'react'
...
+          width="90px"
+          borderColor="darkGrey.100"
+          borderWidth="4px"
+          borderRadius="full"
+        />
+        <HStack>
+          <Heading fontSize="md" fontWeight="700">
+            Full Name
+          </Heading>
+          <Text fontSize="xs" paddingLeft="-2">
```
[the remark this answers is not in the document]
PammyLo: In this screen a mix of sizes is used. The size `xs` is used 4x, `sm` 3x, `md` 2x. 

The `Text` component in `StaTheme` has default size 'md'. If I add fontSize="xs" to the "HStack" and delete the xs props then the font goes bigger overall.  Not sure where else to add a default prop in the theme?

app/src/NativeBase/Containers/ProfileContainer.tsx
PammyLo: Thanks for noticing Joe, it fitted on my screen so didn't realise that. The scrollview should wrap the whole content, not just the wee list. Have fixed :)

app/src/NativeBase/Containers/ProfileContainer.tsx
PammyLo: The card has a corner radius, a different background colour and a border instead of shadow. Could use that but with so many changes it works fine with Box too :)

app/src/NativeBase/Theme/StaTheme.tsx
PammyLo: 🧐

app/src/NativeBase/Containers/ProfileContainer.tsx
joefhall: Ok yeah makes sense, maybe it would be better to change the `Text` default padding in the theme file.  If you're not doing that here, are you ok to create a separate ticket for it so we don't lose this as a task?

app/src/NativeBase/Containers/ProfileContainer.tsx
joefhall: Got it -- makes sense, thanks

joefhall: Another thought - is this box style being used elsewhere in any of the new designs?  And if so, could you make it into a reusable component?

app/src/NativeBase/Containers/ProfileContainer.tsx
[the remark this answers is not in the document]
joefhall: Two thoughts about possible ways could do this:

- Create some custom semantically named custom font sizes in StaTheme under `fonts.sizes` (e.g. `formLabel: 16,` -- that's just an example name, use whatever names makes most sense)  Then you could reference them here as e.g. `<Text fontSize="formLabel"`.  That way if we decide to change the size of e.g. form labels across the app, we could change it one place rather than dozens.

- Or could create reusable React components with the font sizes set there, but that might get a bit tricky if there are lots of different padding and margin settings going on

app/src/NativeBase/Theme/StaTheme.tsx
joefhall: In the new setup with Expo, fonts get defined in [app/App.tsx](https://github.com/Scottish-Tech-Army/Volunteer-app/blob/main/app/App.tsx) and it's named `'Poppins-Regular'` there so I'm pretty sure that's what we should be using -- but was there a reason you changed it?

app/src/NativeBase/Containers/ProfileContainer.tsx
joefhall: Great, thanks -- that's working now for me 👍

app/src/NativeBase/Containers/ProfileContainer.tsx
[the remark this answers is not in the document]
joefhall: (And yep, no worries not to set it on the `HStack` if that doesn't help)

app/src/NativeBase/Containers/ProfileContainer.tsx
```diff
@@ -13,20 +53,111 @@ import React from 'react'
...
+              _dark={{ borderColor: 'secondaryGrey.40' }}
+              borderWidth="4px"
+              borderRadius="full"
+              marginLeft="24px"
+            />
+            <Icon
+              as={MaterialIcons}
+              color="darkGrey.100"
+              _dark={{ color: 'secondaryGrey.40' }}
+              name="lock"
```
joefhall, line 79: What's the padlock icon for in the designs / why is it needed?  (do we need it?)

app/src/NativeBase/Containers/ProfileContainer.tsx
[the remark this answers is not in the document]
joefhall: Actually, ignore my first bullet point above -- I think this would be anti-pattern (it would work, but NativeBase docs don't suggest anything like this -- they stick to standard font size names like `xs`)

I don't feel super comfortable with all the different font sizes and paddings/margins being set on each element -- we should have a design system where we use reusable components (with common settings around sizing and spacing) in most places in the app, so that creating new screens (and updating any settings later) is much easier and more standardised.  

I don't know how easy it is to standardise things with the current designs (e.g. I don't know if Mia is working that way in Figma), so I'll leave it up to you what you think is possible / makes sense here.

app/src/NativeBase/Containers/ProfileContainer.tsx
PammyLo: Yeah, I think it would be better in a separate ticket, due to the refactor that comes with it and not to hold this one up. We can think of other improvements to the theme perhaps to include in that ticket.

app/src/NativeBase/Containers/ProfileContainer.tsx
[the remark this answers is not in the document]
PammyLo: I agree Joe and feel the same way. Last week had a chat with Mia and she uses the descriptions that are in the Typography section of the Design System so we can see how to make that easier to use in our theme. The same could apply to the padding and margins. For this screen I can't see an easy way not to use so many margins on the fly, but we can have a deep dive into making that better as part of the ticket that refactors the `Text` element.

app/src/NativeBase/Theme/StaTheme.tsx
PammyLo: The reason for changing this is that the `fontWeight` prop had no effect on the actual appearance of the fonts. This is resolved by replacing `Poppins-Regular` which is 400 weight, with `Poppins`. I tried this as it matches the suggested font set up in the Native Base docs. 
However, I do notice the error message now too and it's not there when the app is run from `main` so it's obviously not liking it. Not sure how to fix. Will do some digging later today.

PammyLo: That's it fixed now. The error resolved by sticking to `Poppins-Regular` and adding semantic fonts to enable semiBold Bold and Light variations.

app/src/NativeBase/Containers/ProfileContainer.tsx
PammyLo: I have no idea, haha. I did leave that out for that reason, but David quite rightly mentioned it was missing so now I've added it as I reckon it would be easier removed once checked with Mia and Joanna than added.

app/src/Navigators/Main.tsx
PammyLo: Thanks for reviewing David! I left out the padlock as I don't think we need it, but since I had overlooked the edit button, I have now added the padlock as well. We can delete if we don't need it.
I've also made a few more changes based on Joe's input.

app/src/NativeBase/Containers/ProfileContainer.tsx
PammyLo: Many thanks Marcin! Much appreciated you taking the time to review. Hope you see the same menu items now :)

app/src/NativeBase/Containers/ProfileContainer.tsx
PammyLo: I've asked Mia what it is for so will wait for reply. If it is for future version we could also comment it out before merging. 
Same with the edit button. It will be needed for the profile screen David is going to work on next but it still might be better to comment it out before merging. Thoughts?

PammyLo: Mia confirmed the padlock is for the option where the user can choose whether to make the profile public or not. Think you are right and that this will not be part of MVP+1. We can delete or comment out until use. I'm happy either way.

app/src/NativeBase/Containers/ProfileContainer.tsx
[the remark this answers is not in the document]
joefhall: Thanks for looking into it Pam -- that makes sense 👍

app/src/NativeBase/Containers/ProfileContainer.tsx
joefhall: Maybe comment out for now?  That way the work you've done can still be used by someone in the future

app/src/NativeBase/Theme/StaTheme.tsx
joefhall: Nice one -- that's a good solution 👍

joefhall (approved):
Nice one, thanks @PammyLo and for taking all the time on the feedback/revisions 👍 👍 

(Might be good before merging to comment out anything we know won't be used for MVP+1 -- e.g. the padlock -- but will still be useful for the future, but will leave that with you.)

app/src/NativeBase/Containers/ProfileContainer.tsx
PammyLo: I've removed it as it included an added margin to the profile picture to ensure it remained centred. Might be confusing. Also a real chance this feature will look different if and when it get implemented.

app/src/Navigators/Main.tsx
PammyLo: Have removed the padlock again as it is not needed for MVP+1.

DougieWougie (approved):
Nice one @PammyLo !

PR #188, opened 2023-07-20
Pavi ticket 324

pavithiranShanmugakani:
- [Description](#description)
- [Checklist](#checklist)

# Description

- Created code and unit tests for the ticket SVA-324.

Fixes # (issue number)

## Type of change

- [x] This change requires a documentation update

# Testing locally

- To test locally, add the STA_API_KEY of any similar random value in the .env file of api and app folders . 
- To check if the API is successfully tunnelling, check the readme file under the troubleshooting section for instructions.

# Checklist

- [] I have addressed accessibility, if needed


APP_DEVELOPMENT.md
```diff
@@ -82,6 +82,9 @@ There are a few steps to add and use an environment variable using Expo:
 3. In the file in your code where you want to use this variable, add an import statement at the top of the file `import Constants from 'expo-constants'`
 4. In the same file, at the place in the code where you want to use the variable, you can access your variable with e.g. `Constants.expoConfig?.extra?.amazingApiKey` (note: Typescript considers that the value of this could be `undefined`, so you may need to handle that possibility)
 
+### Version
+API version 1.1 are protected by API_KEY, that needs to be configured in `api/.env` and `app/.env` file. The version to be used in the development can be changed by modifying the STA_API_VERSION in the `app/.env` file in the `app` directory.
```
joefhall, line 86: It looks like this is named `API_KEY` in the API and `STA_API_KEY` in the app -- suggest it might be even better if it's standardised so it's the same everywhere (`STA_API_KEY` seems clearest, to differentiate it from other services), this will probably help avoid confusion so devs put the same line in both their .env files

api/app.js
```diff
@@ -2,7 +2,8 @@ const Bugsnag = require('@bugsnag/js');
 const BugsnagPluginExpress = require('@bugsnag/plugin-express');
 const cors = require('cors');
 const express = require('express');
-const routes = require('./routes/index');
+const v1Routes = require('./routes/v1/index');
+const v2Routes = require('./routes/v1.1/index');
```
joefhall, line 6: For consistency please can this be called `v1_1Routes` instead of `v2Routes` (also at some point we will probably have v2 routes and so the current naming here would be problem)

api/__tests__/routes/apiKeyMiddleware.test.js
```diff
@@ -0,0 +1,61 @@
+const axios = require('axios');
+const eventsHelper = require('../../helpers/events');
+const { apiKeyMiddleware } = require('../../routes/apiKeyMiddleware');
+axios.defaults.adapter = require('axios/lib/adapters/http');
+process.env = {
```
joefhall, line 5: There's a risk with this that it could mess up other tests that also rely on `process.env` as you're altering the value here but not resetting it after the test.  To avoid that:

- Move this line here (where you set `process.env`) into inside each of your two tests below
- Add lines 7-16 from [this file](https://github.com/Scottish-Tech-Army/Volunteer-app/blob/develop/api/__tests__/services/logging.test.js#L7) after the `describe(...` line below and before the first `test(...` -- this will reset `process.env` before/after your tests

joefhall (changes requested):
Nice work @pavithiranShanmugakani, looking good!

Have just added a few comments.

A couple of other things I've thought of:
- In the README there are a few references to checking v1 URLs in the browser (search for v1 in that file to find them) -- these should be changed to v1.1 and people should use Postman with STA_API_KEY in the header rather than using their browser.  It would be nice if you could link to somewhere that explains how to do that as some folks may be entirely new to Postman.
- Similarly, [on this line](https://github.com/Scottish-Tech-Army/Volunteer-app/blob/b4c401d22dcfd71d878b801c7544beccaebf20d7/app/App.tsx#L85) in App.tsx that needs to be changed as it currently tells people to try it in the browser

For others who are reviewing, would be good if you could fill in the 'Testing locally' section of your pull request description (remove the text that's already there) so people know how to test it out locally (e.g. what do they need to update in their .env files, how to check if it's working)

joefhall:
@pavithiranShanmugakani I also spotted it looks like this was branched from an outdated version of some of the code so your branch needs updating with the latest changes from `main` or `develop`.  The API tunnelling solution now uses localtunnel, it would be good to check everything still works after you've done that.

app/src/Services/modules/projects/registerInterest.ts
DJMF2015, line 1: Good work and it works fine on my local machine. Like the additonal clarity around the ReadMe and more good practices surrounding the versioning of the middleware. Happy to approve.

DJMF2015 (approved):
Nice work on the Readme and versioning of the middleware and routers. Happy to approve.

README.md
```diff
@@ -172,9 +172,16 @@ Below are some commonly encountered issues and possible ways to resolve them. If
...
-
-  > Once a week, your API tunnel will expire and you'll need to enter an IP address in order to continue using it. To check if this is the case, go to the URL of your tunnelled API and add `/v1/projects` at the end (e.g. `https://nadia-bloggs-abc123def456ghi789.loca.lt/v1/projects`) If you see a web page instead of projects data, follow the instructions on the page to enter your public IP address -- follow what it says in the *If you're the developer...* section further down the page. Then stop and restart your app in Expo.
+ > ### Using Postman
+ > Install postman by [Link](https://www.postman.com/downloads/). Once installed and running, select 'import' in the app and use the following cURL code with the api key pasted after the colon. The cURL will automatically build the postman request for you. To check if the api key works, toggle and untoggle the key and send your request. Some examples of the api key in the header is found in this [link](https://blog.postman.com/how-to-use-api-keys/) for extra knowledge.
+ > ```cURL
```
CatMac228, line 177: Teeny tiny suggestion! I had to select 'Raw Text' after import to paste the cURL code and it might be worth adding this into the instructions. I was selecting the cURL option and was confused when it was expecting a file to be added :)

pavithiranShanmugakani: Hi Cat. I have addressed the changes to the readme. Please have a look. many thanks.

CatMac228 (approved):

PR #192, opened 2023-08-01
Sva 293 add es lint and prettier config to api directory

M4RC1N76:
- [Description](#description)
 

# Description

- Added a prettierrc.json to the /api/ directory
- Added a eslintrc.js config file to the /api/ directory
- Have converted CommonJS require statements into ES6 import statements
- Have verified the code re-formats upon save with Prettier formatting, and when running npx eslint .  the linter runs correctly and provides errors. There will need to be a separate ticket to fix all issues in the linters, and some of the config rules might need updating depending on this.
- In the /api/ directory package.json, i've installed two additional dependencies, as I was having issues with eslint and prettier. Installing the plugin 'eslint-plugin-prettier' fixed these issues. (I suspect it is because i'm using Prettier v3.0, unlike the app directory which uses Prettier v2.8.8)
- Fixes # (SVA-293)

## Type of change

# Testing locally

- Run `npm install` and ensure that Prettier works and run the command `npx eslint .` in the api directory to ensure that you can see the linting working. You should see 158 problems in the results. This is expected and need to be addressed in a separate ticket.

# Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have addressed accessibility, if needed
- [ ] I have checked the app in dark mode, if making front-end design changes
- [ ] My changes generate no new warnings
- [ ] I have updated the version numbers in `package.json` files in the [app](DEPLOYMENT.md#app-deployment) and/or [api](DEPLOYMENT.md#api-deployment-on-aws) directories as needed

### DIFF
--- a/api/cron-jobs/projects.js
+++ b/api/cron-jobs/projects.js
@@ -1,46 +1,46 @@
-require('dotenv').config();
-const airTable = require('../helpers/airTable');
-const arraysHelpers = require('../helpers/arrays');
-const axios = require('axios').default;
-const urlsHelpers = require('../helpers/urls');
-const projectsHelpers = require('../helpers/projects');
-const videosService = require('../services/videos');
-const logging = require('../services/logging');
-const api_key = process.env.JIRA_API_KEY;
-const email = process.env.JIRA_EMAIL;
-const resourcingJiraBoardName = 'RES';
-const recruiterAssignedJiraColumnName = 'Recruiter Assigned';
-const projectJiraBoardName = 'IT';
-const volunteerSearch = 'Volunteer Search';
-const volunteerIntroduction = 'Volunteer Introduction';
-const activityUnderway = 'Activity Underway';
+require('dotenv').config()
+const airTable = require('../helpers/airTable')
+const arraysHelpers = require('../helpers/arrays')
+const axios = require('axios').default
+const urlsHelpers = require('../helpers/urls')
+const projectsHelpers = require('../helpers/projects')
+const videosService = require('../services/videos')
+const logging = require('../services/logging')
+const api_key = process.env.JIRA_API_KEY
+const email = process.env.JIRA_EMAIL
+const resourcingJiraBoardName = 'RES'
+const recruiterAssignedJiraColumnName = 'Recruiter Assigned'
+const projectJiraBoardName = 'IT'
+const volunteerSearch = 'Volunteer Search'
+const volunteerIntroduction = 'Volunteer Introduction'
+const activityUnderway = 'Activity Underway'
 
 async function addNewProjectsResources(projectsResources) {
   // AirTable accepts creating records in groups of 10 (faster than doing just one record at at time),
   // so we chunk our data into an array of arrays, where each top-level array item is an array of up to 10 projects/resources
 
-  const projectsResourcesChunked = arraysHelpers.chunk(projectsResources, 10);
+  const projectsResourcesChunked = arraysHelpers.chunk(projectsResources, 10)
 
-  console.log('🛈 Saving project resources');
+  console.log('🛈 Saving project resources')
   for (const projectsResourcesChunk of projectsResourcesChunked) {
     await module.exports.addNewRecords(
       airTable.projectsResourcesCacheTable(),
-      projectsResourcesChunk
-    );
+      projectsResourcesChunk,
+    )
   }
 
-  console.log('🛈 Finished saving new cache data');
+  console.log('🛈 Finished saving new cache data')
 }
 
 async function addNewRecords(tableName, recordsChunk) {
-  const recordsChunkFormattedForAirTable = recordsChunk.map((record) => ({
+  const recordsChunkFormattedForAirTable = recordsChunk.map(record => ({
     fields: record,
-  }));
+  }))
 
   await airTable
     .client()
     .table(tableName)
-    .create(recordsChunkFormattedForAirTable);
+    .create(recordsChunkFormattedForAirTable)
 }
 
 async function cacheProjectsAndResources(projects, resources) {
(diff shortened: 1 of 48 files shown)


kelsiesmurphy:
Commenting so i can get notified and answer any questions on the bits I did :)

api/cron-jobs/projects.js
```diff
@@ -1,332 +1,305 @@
...
-const logging = require('../services/logging');
-const api_key = process.env.JIRA_API_KEY;
-const email = process.env.JIRA_EMAIL;
-const resourcingJiraBoardName = 'RES';
-const recruiterAssignedJiraColumnName = 'Recruiter Assigned';
-const projectJiraBoardName = 'IT';
-const volunteerSearch = 'Volunteer Search';
-const volunteerIntroduction = 'Volunteer Introduction';
-const activityUnderway = 'Activity Underway';
+require('dotenv').config()
```
joefhall, line 1: Same point as before that I think this can be updated to an `import` (please see earlier comment)

api/cron-jobs/projects.js
```diff
@@ -1,332 +1,305 @@
...
+  const projectsResourcesChunked = chunk(projectsResources, 10)
 
-  console.log('🛈 Saving project resources');
+  console.log('🛈 Saving project resources')
   for (const projectsResourcesChunk of projectsResourcesChunked) {
-    await module.exports.addNewRecords(
-      airTable.projectsResourcesCacheTable(),
-      projectsResourcesChunk
-    );
+    await _addNewRecords(projectsResourcesCacheTable(), projectsResourcesChunk)
```
joefhall, line 29: I didn't know about this underscore syntax, that's good to learn! 🔥

api/helpers/airTable.js
```diff
@@ -1,116 +1,122 @@
-require('dotenv').config();
-const AirTable = require('airtable');
-const dayjs = require('dayjs');
-const relativeTime = require('dayjs/plugin/relativeTime');
-dayjs.extend(relativeTime);
-const logging = require('../services/logging');
+require('dotenv').config()
```
joefhall, line 1: As above -- please see earlier comment

api/package.json
```diff
@@ -32,9 +36,18 @@
...
     "cross-var": "^1.1.0",
     "dotenv-cli": "^7.2.1",
+    "eslint-plugin-prettier": "^5.0.0",
     "jest": "^27.2.4",
     "nock": "^13.2.0",
     "superagent": "^6.1.0",
     "supertest": "^6.1.6"
+  },
+  "lint-staged": {
+    "*.{js,ts,tsx, jsx}": [
```
joefhall, line 46: I don't think we should be including `tsx` or `jsx` extensions here, as those are specifically for React

api/package.json
```diff
@@ -32,9 +36,18 @@
...
+      "eslint --fix"
+    ],
+    "*.{json,md,html,js,jsx,ts,tsx}": [
```
joefhall, line 49: As above about `tsx` and `jsx`

api/services/slack.js
```diff
@@ -4,37 +4,37 @@
  * See instructions in the README on how to enable new Slack channels
  */
 
-require('dotenv').config();
-const axios = require('axios').default;
-const logging = require('../services/logging');
+require('dotenv').config()
```
joefhall, line 7: As above - please see other comment

joefhall (changes requested):
Epic work @M4RC1N76 @kelsiesmurphy -- thanks a lot!

Have scanned through everything and looking good but can't guarantee I haven't missed anything, as obviously it's an enormous diff.

Have you been able to a) get the API running locally with these changes and b) get tests to pass?  Both are failing for me at the moment, though possible it's just small fixes that are needed to get these working.

Also spotted a few small things, please see questions/comments.

joefhall:
Hey @kelsiesmurphy @M4RC1N76 just to flag, I've merged my latest PR into `develop` now so you'll need to merge those latest changes into your PR here.  (Obviously no need to worry about the changes that were just formatting, but the updated functionality and tests are needed.)

joefhall:
Hi again @kelsiesmurphy @M4RC1N76 I think there's a solution to the problem you were having with imports.  Looking at this answer https://stackoverflow.com/a/59399717/15422951 

If you do option 1 you need to:
- add `"type": "module",` to api/package.json (at the top level below e.g. `"main": "app.js",`) -- this tells Node to treat files as modules by default
- all the import statements need to use `.js` in their filenames (e.g. `import { listen } from './app.js'`)

Alternatively you could do their option 2 and rename all the JS module files to .mjs which tells Node they are modules

I'm not certain which is easier.  Option 1 feels a bit neater?  Option 2 might be a quicker job?  Not sure what option 2 would do to the diff (hopefully GitHub will just see it as a filename change, not show the whole file in the PR?)  Option 2 might also work better if there are any JS files which shouldn't be treated as modules -- or is that all of them now anyway (including tests)?  If it is all of  them maybe option 1 is better.

kelsiesmurphy:
> Hi again @kelsiesmurphy @M4RC1N76 I think there's a solution to the problem you were having with imports. Looking at this answer https://stackoverflow.com/a/59399717/15422951
> 
> If you do option 1 you need to:
> 
> * add `"type": "module",` to api/package.json (at the top level below e.g. `"main": "app.js",`) -- this tells Node to treat files as modules by default
> * all the import statements need to use `.js` in their filenames (e.g. `import { listen } from './app.js'`)
> 
> Alternatively you could do their option 2 and rename all the JS module files to .mjs which tells Node they are modules
> 
> I'm not certain which is easier. Option 1 feels a bit neater? Option 2 might be a quicker job? Not sure what option 2 would do to the diff (hopefully GitHub will just see it as a filename change, not show the whole file in the PR?) Option 2 might also work better if there are any JS files which shouldn't be treated as modules -- or is that all of them now anyway (including tests)? If it is all of them maybe option 1 is better.

So I had already went down this path, and I started getting heaps of issues surrounding 'export default X' and the equivalent destructured 'import {X} from Y' routes. Some of the particular issues were around importing JSON into a file, which we do several times, and i think I started going around in circles with it. When I have an hour or two free i'll revisit it :)

joefhall:
Good luck with it!  It's pesky -- if we were using Typescript, it would be super simple _(one day...)_

You may well have seen this already, but just in case not: https://www.stefanjudis.com/snippets/how-to-import-json-files-in-es-modules-node-js/

PR #197, opened 2023-09-29
SVA-387 Update theme components and refactor code to match

PammyLo:
- [Description](#description)
- [Checklist](#checklist)

# Description

- Throughout the app many different font sizes and paddings/margins were being set. This makes some changes to a few of the Theme components to standardise and avoid need for corrective margins or negative padding. 
   - Biggest impact requiring some margin and padding changes on all screens was removing the default padding from the `Text` component
   - Other change worth noting is introducing a way to easily set the font style following the naming in the Design System in Figma. 
   - Refactored codebase to ensure layout is preserved
 
- Overall there should now be less margins and paddings set throughout the codebase and no more need for negative padding. 

- Also updated the look of the date picker which was missing a few border lines.  

Fixes # (issue number)
[SVA-387](https://sta2020.atlassian.net/browse/SVA-387)

## Type of change

- [X] Non-breaking change which fixes an issue
- [] New Feature (non-breaking change which adds functionality)

# Testing locally

- install npm before starting the app
- start as normal and check the app still looks good 
    - the only pages changed are fixed margins in Profile container and improved look of the calendar date picker (in Register screen)

# Checklist

- [x] I have updated the version numbers in `package.json` files in the [app](DEPLOYMENT.md#app-deployment) and/or [api](DEPLOYMENT.md#api-deployment-on-aws) directories as needed

### DIFF
--- a/app/src/NativeBase/Components/Project/ProjectFullDetails.tsx
+++ b/app/src/NativeBase/Components/Project/ProjectFullDetails.tsx
@@ -11,6 +11,7 @@ import {
   Card,
   Heading,
   HStack,
+  Icon,
   ScrollView,
   Text,
   VStack,
@@ -19,6 +20,7 @@ import Video from '../Video'
 import ProjectAttachments from '../ProjectAttachments'
 import ColouredTag from '../ColouredTag'
 import StaTheme from '../../Theme/StaTheme'
+import MaterialIcons from 'react-native-vector-icons/MaterialIcons'
 
 interface ProjectFullDetailsProps {
   project: Project
@@ -41,8 +43,11 @@ const ProjectFullDetails = ({
             justifyContent="space-between"
             alignItems="center"
             paddingRight="4"
+            space={4}
           >
-            <Heading fontSize="md">{project.name}</Heading>
+            <Heading fontSize="md" width="70%">
+              {project.name}
+            </Heading>
 
             {/* TODO: reinstate when functionality is ready and set Heading width="70%" */}
             {/* <HStack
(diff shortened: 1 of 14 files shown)


M4RC1N76 (approved):
Great work @PammyLo Pamela, I've familiarise myself with the changes and run the App on my android. All good!!

app/src/NativeBase/Components/Project/ProjectFullDetails.tsx
DJMF2015, line 1: So far the work is great and good to see it's now possible to 'do away' with negative margining and there are the set paddings and margins laid specified. I also like the fact that you have tidied up some of the layouts to remove unnecessary components e.g.) multiple VStacks. I just have one quick question. I noticed you've added in the 'react-native-vector-icons/MaterialIcons' import from Material Icons. Is this installed as an additional dependency? If so, is it possible to achieve the same result with existing icon libraries installed? For instance 'expo-vector-icons'? I mention this as when i did something similar before the break, Joe was adamant we shouldn't be adding additional dependencies in if at all possible for consistency. ie) we should be using existing expo libraries where possible. I would be keen to know your thoughts! 😊

app/src/NativeBase/Theme/StaTheme.tsx
DJMF2015, line 1: Good work!

app/src/NativeBase/Components/Project/ProjectFullDetails.tsx
PammyLo: Hi David, thank you so much for reviewing! You are absolutely right and I agree totally. To put your mind at ease, this is an existing package, it is only imported into this file to use an icon from it. Thanks for keeping us right! :)

DJMF2015: That's great. Thanks for clarifying this!

DJMF2015 (approved):
The code changes and amendments are looking good and i am happy to approve your PR. Good work.

PR #200, opened 2023-10-16
SVA-403 Adds authentication API

roryf:
# Description

https://sta2020.atlassian.net/browse/SVA-403

Adds `/auth` endpoints to support Cognito-based authentication in the app.

Requires config updates to set Cognito env vars, depends on setting up Cognito correctly.

## Type of change

Please delete options that are not relevant

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New Feature (non-breaking change which adds functionality)
- [ ] Breaking change(fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# Testing locally

Won't currently work locally or otherwise...

# Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have removed any unnecessary comments or console logging
- [ ] I have made corresponding changes to the documentation (if required)
- [ ] I have addressed accessibility, if needed
- [ ] I have followed best practices, e.g. NativeBase approaches and theming
- [ ] I have checked the app in dark mode, if making front-end design changes
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have updated the version numbers in `package.json` files in the [app](DEPLOYMENT.md#app-deployment) and/or [api](DEPLOYMENT.md#api-deployment-on-aws) directories as needed

PR #203, opened 2023-10-20
SVA-409 Adds authentication to app

roryf:
# Description

Authentication services within app.

Adds `AuthProvider` which is configured at app level and can be consumed in any component via `useAuth` hook. Added a placeholder example of detecting auth state in `ProfileContainer`

There is an underlying `AuthClient` class which handles specific API interaction and storage of auth credentials, which uses [`expo-secure-store`](https://docs.expo.dev/versions/latest/sdk/securestore/). This also manages automatically refreshing auth tokens when access token expires, and handling transient network errors (which will happen a lot).

Also configured jest correctly so we can add tests, and removed unused `app/__tests__` directory (I strongly prefer test files to sit with the file under test). Rejigged the app CI build too so it runs the tests, and lints all files not just changed (not convinced this worked correctly tbh). I would very much like to revisit the different eslintrc configs at some point, and the need for questionably useful (IMHO) file comments.

Depends on backend in #200 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change(fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# Testing locally

No user-facing features added, run the app, check nothing breaks and there are no new console warnings.

# Checklist

- [] I have performed a self-review of my own code
- [] I have commented my code, particularly in hard-to-understand areas
- [] I have removed any unnecessary comments or console logging
- [] I have addressed accessibility, if needed
- [] I have followed best practices, e.g. NativeBase approaches and theming
- [] My changes generate no new warnings


app/src/Services/auth/AuthProvider.test.tsx
```diff
@@ -0,0 +1,272 @@
...
+      )
+      const { queryByText } = render(<Text>Authentication test</Text>, {
+        wrapper: blockingWrapper,
+      })
+
+      expect(queryByText('Authentication test')).toBeNull()
+      await waitFor(() =>
+        expect(queryByText('Authentication test')).toBeTruthy(),
+      )
+    }, 10000) // Without this, the test fails on GitHub Actions run.
```
roryf, line 269: I don't know why this is, the test covers a simple `useEffect` hook with an async function that should resolve immediately on next tick. It passes locally without issue, and I have used this technique before in other codebases with any CI issues. I'm comfortable leaving it in place for now until someone has time to properly investigate.

DJMF2015 (approved):
@roryf I have run the app and tested locally and the app runs fine with no new warnings generated. I have looked over the code and it looks great work. I saw your remark about the async function though i am not too sure of the cause either. One thing i noticed when i run the app is that i receive a a couple of `console.logs `in the command line stating " No ID token in storage getIdToken" and "No ID token found in getCurrentUser". I take it these are just logged from the testcases which are generating it? Taht aside, all test cases pass for me and no issues. Happy to approve and good work.

roryf:
Thanks @DJMF2015, the console.logs are actually deliberate and for debugging purposes, when I've used similar auth solutions to this before with an error tracking tool it was incredibly useful to see the state of the auth client, so I would be inclined to leave them.

DJMF2015:
> Thanks @DJMF2015, the console.logs are actually deliberate and for debugging purposes, when I've used similar auth solutions to this before with an error tracking tool it was incredibly useful to see the state of the auth client, so I would be inclined to leave them.

M4RC1N76:
I am testing the 403 on the Iphone emulator on my mac, the Apps works fine. I have noticed a few console log's in the terminal on the app side regarding https address I am using (and probability of it not working, there is also a suggestion to use the same https address in postman). Also there is a console log about me having a large list (VirtualizedList) which is slow to update (also received some hints: to "make sure your renderItem function renders components that follow React performance best practices like PureComponent, shouldComponentUpdate, etc.") I hope the description helps. As mentioned above the App works!! Is there something else I should do or this is sufficient for the PR approval? @roryf let me know what you think, please?

roryf:
Thanks @M4RC1N76, all of the issues you mentioned were existing so I think that's good enough!

M4RC1N76 (approved):
Thank you for the clarification @roryf , great work!! PR approved.

CatMac228 (approved):

PR #219, opened 2024-11-21
SVA-444 update filters to new dropdown designs

mirkiy:
Jira ticket: SVA-444- Roles List - update filters to new drop down designs (curently called projects list)


src/NativeBase/Components/TagButtons.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+const TagButtons: React.FC<TagButtonsProps> = ({
+  tags,
+  iconState,
+  handleTagPress,
+}) => {
+  return (
+    <HStack space={2} alignItems="center">
+      {tags.map(tag => (
+        <Pressable
```
dkaschl, line 19: I'd also extract a single tagbutton into it's own component. This component here (TagButtons) could be then just using those tagbutton components. Naming wise we could also name HTagButtons (H for horinzontal) but as we do have only one version currently I think FilterTagButtons would be a good name as it says that it's about tag buttons used for filtering.

src/NativeBase/Containers/ListContainer.tsx
```diff
@@ -269,77 +318,143 @@ const ListContainer = (props: {
...
                   />
-                </SearchResultsView>
-              )}
+                }
+              />
+            </View>
+          </HStack>
+
+          {/* Display filtered roles based on the search */}
+          <SearchResultsView>
```
dkaschl, line 371: I am not exactly sure how the search result should be implemented, but when it comes to results I suppose it should be the same style like the other lists and not in this section here. Also searching after every keychange is performancewise a bit costly (and makes the system also slow) so I'd advise here to use something like a debouncing pattern.

src/NativeBase/Containers/ListContainer.tsx
```diff
@@ -269,77 +318,143 @@ const ListContainer = (props: {
...
+            )}
+            {roleSearchText && filteredRoles.length === 0 && (
+              <Text>No roles found</Text>
+            )}
+          </SearchResultsView>
+        </VStack>
+
+        <VStack>
+          <ScrollView horizontal showsHorizontalScrollIndicator={false}>
+            <ProjectSearchContainer />
```
dkaschl, line 392: we should rethink about the structure and/or the naming of the components. In this component we put together the freesearchbar with it's extra icon on the right "unwrapped" but then wrapp the tagbuttons with it's choicelists in "ProjectSearchContainer" which is actually not the correct naming as it's actually filtering.

src/NativeBase/Containers/ProjectSearchContainer.tsx
dkaschl, line 1: I've added comments on how this component is used in ListContainer.tsx. I think we should rename it and maybe restructure it too.

src/Navigators/Main.tsx
```diff
@@ -7,7 +7,7 @@ import {
   BottomTabNavigationOptions,
 } from '@react-navigation/bottom-tabs'
 import { useRoute } from '@react-navigation/native'
-import { ListContainer } from '@/NativeBase/Containers'
+import { ListContainer, ProjectSearchContainer } from '@/NativeBase/Containers'
```
dkaschl, line 10: please remove unused imports

src/NativeBase/Containers/ProjectSearchContainer.tsx
```diff
@@ -1,66 +1,43 @@
...
+import { searchByArray, fuzzySearchByArray } from '@/Utils/Search'
 import {
-  roleGroups,
   RoleGroupName,
+  roleGroups,
 } from '@/Services/modules/projects/roleGroups'
-import { ProjectsState } from '@/Store/Projects'
-import { searchByArray, fuzzySearchByArray } from '@/Utils/Search'
-
-enum Tab {
```
dkaschl, line 33: i would have kept the enum, as we would use only the enums throughout this code and if we want to change a name to another one we'd only change it once here.

dkaschl (changes requested):
Give the rather unclear scope and explanation of the ticket mixed with a design where I'm not sure which version we really need to implement, it's a pretty complex and difficult task! As already mentioned it would be important to clarify the exact scope of this ticket and also address what the UI should really be in order to know what exactly what we need to do :)

